PR #10, opened 2026-04-02
Add Ollama LLM adapter and support

Masteromanlol:
Introduce a new OllamaAdapter (src/llm/ollama.ts) implementing LLMAdapter with chat and streaming support, converting between the framework's ContentBlock types and Ollama (OpenAI-compatible) chat completions. Wire the adapter into the factory (src/llm/adapter.ts) and extend provider types (src/types.ts) to include 'ollama'. Update example (examples/04-multi-model-team.ts) to allow selecting Ollama as a model/provider option. Ollama adapter defaults its base URL from OLLAMA_BASE_URL or http://localhost:11434 and handles tool calls, tool results, images, and finish reason normalization.

### DIFF
--- a/examples/04-multi-model-team.ts
+++ b/examples/04-multi-model-team.ts
@@ -151,8 +151,7 @@ Return the raw rates as a JSON object keyed by pair, e.g. { "USD/EUR": 0.91, "US
 
 const analystConfig: AgentConfig = {
   name: 'analyst',
-  model: useOpenAI ? 'gpt-5.4' : 'claude-sonnet-4-6',
-  provider: useOpenAI ? 'openai' : 'anthropic',
+  model: useOllama ? 'llama3.1' : useOpenAI ? 'gpt-4o-mini' : 'claude-3-5-sonnet-20240620',\n  provider: useOllama ? 'ollama' : useOpenAI ? 'openai' : 'anthropic',
   systemPrompt: `You are a foreign exchange analyst.
 You receive exchange rate data and produce a short briefing.
 Use format_currency to show example conversions.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1459,6 +1459,12 @@
         }
       }
     },
+    "node_modules/ollama": {
+      "version": "0.3.0",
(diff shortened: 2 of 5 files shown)


JackChen-me:
Thanks for adding Ollama support — local model inference is a great addition to the framework, and the adapter design follows the right pattern. Appreciate the contribution!

Before merging, there are a few issues that need to be addressed:

1. Merge conflict with main

This PR conflicts with the recent Copilot adapter merge (PR #9). A rebase onto current main is needed.

2. Reuse openai-common.ts instead of duplicating helpers

PR #9 extracted shared OpenAI wire-format helpers (toOpenAITool, toOpenAIMessages, fromOpenAICompletion, normalizeFinishReason, buildOpenAIMessageList) into src/llm/openai-common.ts. The Ollama adapter should import from there — same as openai.ts and copilot.ts do — rather than copying ~180 lines of identical conversion logic.

3. readonly and indentation lost in types.ts

The provider field in AgentConfig and defaultProvider in OrchestratorConfig lost their readonly modifier and leading indentation. Also needs to include 'copilot' from main. Should be:

readonly provider?: 'anthropic' | 'copilot' | 'ollama' | 'openai'

4. Syntax error in example file

In examples/04-multi-model-team.ts, the model and provider lines are joined with a literal \n, which is a TypeScript syntax error. Also, useOllama is referenced but never defined.

5. OpenAI SDK client initialization

Two issues in the constructor:

No apiKey provided — the OpenAI SDK will throw if OPENAI_API_KEY is unset. Pass a dummy value like apiKey: 'ollama'.
baseURL should include /v1 (Ollama's OpenAI-compatible endpoint is at /v1/chat/completions). Default to http://localhost:11434/v1.

6. Phantom ollama npm dependency

package-lock.json includes the ollama package, but it's not in package.json and the code doesn't use it (it uses the openai SDK). This entry should be removed.

For reference, PR #9 went through a similar review cycle and the author addressed all 7 comments in a follow-up commit before merge. Happy to re-review once these are fixed!

Let me know if you need any help with the rebase or have questions — happy to assist.

PR #48, opened 2026-04-05
Feat/redis store

baiyuqing:
## Checklist

- [ ] `npm run lint` passes
- [ ] `npm test` passes
- [ ] No new runtime dependencies (or justified in the PR description)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,59 @@
+# AGENTS.md
+
+## Commands
+
+```bash
+npm run build          # tsc, src/ → dist/
+npm run dev            # tsc --watch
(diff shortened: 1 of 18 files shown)


JackChen-me:
Saw this closed fast with no comment. Guessing it was because the README lists persistence as a non-goal, but your design (optional peer dep, in-memory default, backward-compatible constructors) doesn't actually break the three-dependency promise. That promise is about runtime deps, and we already have two optional peers (`@google/genai`, `@modelcontextprotocol/sdk`). 

Want to reopen? The design looks solid enough that I want to review it before deciding.

PR #88, opened 2026-04-09
feat: Add contextStrategy to control conversation growth and prevent token explosion in agent runs

### LINKED ISSUE #59: [P1] Context Window Management — sliding window and summarization for long conversations
## Context

When an agent runs multiple tool-call turns, the conversation `messages` array grows without bound. Each turn adds an assistant message (with tool-use blocks) and a user message (with tool results). A 10-turn agent run with file reads can easily accumulate 50K+ tokens of conversation history, which:

1. **Exceeds model context windows** — local models (8K-32K context) hit limits fast; even Claude and GPT slow down and degrade with very long contexts
2. **Inflates cost** — every LLM call resends the entire conversation; input token costs grow quadratically with turn count
3. **Degrades quality** — models perform worse on tasks buried in long context ("lost in the middle" effect)

Today, the only controls are `maxTurns` (hard stop) and `maxTokens` (output cap). There is no mechanism to manage **input context size** during a conversation. The `AgentRunner.stream()` loop (`runner.ts:228`) pushes every message into `conversationMessages` and sends all of them on every LLM call (`runner.ts:285`), with no truncation or compression.

This is the framework's most pressing production-readiness gap for long-running agent tasks.

### New `contextStrategy` option on `AgentConfig`

```typescript
const agent: AgentConfig = {
  name: 'researcher',
  model: 'claude-sonnet-4-6',
  systemPrompt: 'You research topics thoroughly.',
  maxTurns: 30,

  // Option A: keep only the last N turns
  contextStrategy: { type: 'sliding-window', maxTurns: 10 },

  // Option B: summarize old messages when context gets large
  contextStrategy: { type: 'summarize', maxTokens: 50000 },

  // Option C: custom callback
  contextStrategy: {
    type: 'custom',
    compress: async (messages, tokenCount) => {
      // Return a compressed messages array
      return messages.slice(-10)
    },
  },
}
```

### Type definitions

```typescript
/** Context management strategy for long conversations. */
export type ContextStrategy =
  | { type: 'sliding-window'; maxTurns: number }
  | { type: 'summarize'; maxTokens: number; summaryModel?: string }
  | { type: 'custom'; compress: (messages: LLMMessage[], estimatedTokens: number) => Promise<LLMMessage[]> | LLMMessage[] }
```

### Integration point: `AgentRunner.stream()` (`runner.ts:268`)

Insert a context management step before the LLM call in the main loop:

```typescript
// runner.ts — inside the while(true) loop, before step 1

// Context management: compress conversation if strategy is configured
if (this.options.contextStrategy && turns > 1) {
  conversationMessages = await this.applyContextStrategy(
    conversationMessages,
    this.options.contextStrategy,
  )
}

// Step 1: Call the LLM
const response = await this.adapter.chat(conversationMessages, baseChatOptions)
```

### Strategy implementations

**Sliding window** (simplest, no extra LLM calls):
- Count message pairs (assistant + user = 1 turn)
- When turn count exceeds `maxTurns`, drop the oldest turns
- Always preserve: the first user message (original prompt) and the system prompt
- Optionally prepend a note: `"[Earlier conversation history truncated — {N} turns removed]"`

**Summarize** (higher quality, costs one extra LLM call):
- Estimate token count using a simple heuristic (chars / 4 for English)
- When estimated tokens exceed `maxTokens`, split messages into "old" and "recent" halves
- Summarize "old" messages with a single LLM call: `"Summarize the following conversation history..."`
- Replace old messages with a single user message containing the summary
- Use `summaryModel` if specified (allows using a cheaper/faster model for summaries)
- Cache: don't re-summarize if the old portion hasn't changed since last summarization

**Custom** (escape hatch):
- Call the user's `compress` function with the full message array and estimated token count
- Use whatever the callback returns as the new messages array
- Validate: result must be a valid `LLMMessage[]` with at least one message

### Token estimation

Add a lightweight utility (no tokenizer dependency):

```typescript
// utils/tokens.ts
export function estimateTokens(messages: LLMMessage[]): number {
  let chars = 0
  for (const msg of messages) {
    for (const block of msg.content) {
      if (block.type === 'text') chars += block.text.length
      else if (block.type === 'tool_result') chars += block.content.length
      else if (block.type === 'tool_use') chars += JSON.stringify(block.input).length
    }
  }
  // ~4 chars per token for English text (conservative estimate)
  return Math.ceil(chars / 4)
}
```

### Key decisions

- **Runner-level, not Agent-level** — strategy is applied inside `AgentRunner.stream()` so it works for both `run()` and `stream()`. Config is passed through `RunnerOptions`.
- **No tokenizer dependency** — use char-based estimation. Exact token counts require model-specific tokenizers which would add heavy dependencies. The estimation is "good enough" for triggering compression.
- **Preserve first message** — the original user prompt is always kept, even in sliding-window mode. Losing the original intent degrades output quality more than any other single message.
- **Summarize uses the same adapter** — reuses the agent's own `LLMAdapter` for summary calls (or a separate adapter if `summaryModel` is specified). No new infrastructure needed.
- **Opt-in** — zero behavior change when `contextStrategy` is not set. Existing code is unaffected.

## Acceptance Criteria

- [ ] `ContextStrategy` type added to `types.ts`
- [ ] `contextStrategy` optional field on `AgentConfig` (threaded through to `RunnerOptions`)
- [ ] `sliding-window` strategy: drops oldest turns, preserves first user message
- [ ] `summarize` strategy: compresses old messages via LLM call, respects `maxTokens` threshold
- [ ] `custom` strategy: calls user-provided callback
- [ ] `estimateTokens()` utility in `utils/tokens.ts`
- [ ] Context management runs before each LLM call in the main loop
- [ ] `onTrace` emits a trace event for summarization LLM calls (distinguishable from regular turns)
- [ ] Tests cover: sliding-window truncation, summarize trigger and output, custom callback, no-strategy baseline, first-message preservation
- [ ] Example added or existing example updated to demonstrate `contextStrategy`

ibrahimkzmv:
## What

Introduces a `contextStrategy` option to manage conversation history during multi-turn agent runs. Supports sliding-window truncation, LLM-based summarization, and custom compression to prevent unbounded context growth.

## Why

Currently, agent conversations grow without limit, causing:
- Context window overflows (especially for smaller/local models)
- Rapidly increasing token costs due to full history being resent each turn
- Degraded model performance from excessively long context

This change ensures context size is actively managed during execution, making long-running agent workflows production-ready and cost-efficient.

Closes #59 

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/examples/01-single-agent.ts
+++ b/examples/01-single-agent.ts
@@ -114,6 +114,8 @@ const conversationAgent = new Agent(
     model: 'claude-sonnet-4-6',
     systemPrompt: 'You are a TypeScript tutor. Give short, direct answers.',
     maxTurns: 2,
+    // Keep only the most recent turn in long prompt() conversations.
+    contextStrategy: { type: 'sliding-window', maxTurns: 1 },
   },
   new ToolRegistry(), // no tools needed for this conversation
   new ToolExecutor(new ToolRegistry()),
--- a/src/agent/agent.ts
+++ b/src/agent/agent.ts
@@ -153,6 +153,7 @@ export class Agent {
       agentRole: this.config.systemPrompt?.slice(0, 50) ?? 'assistant',
       loopDetection: this.config.loopDetection,
       maxTokenBudget: this.config.maxTokenBudget,
+      contextStrategy: this.config.contextStrategy,
     }
 
     this.runner = new AgentRunner(
--- a/src/agent/runner.ts
+++ b/src/agent/runner.ts
@@ -29,10 +29,12 @@ import type {
   LoopDetectionConfig,
   LoopDetectionInfo,
   LLMToolDef,
+  ContextStrategy,
 } from '../types.js'
 import { TokenBudgetExceededError } from '../errors.js'
 import { LoopDetector } from './loop-detector.js'
 import { emitTrace } from '../utils/trace.js'
+import { estimateTokens } from '../utils/tokens.js'
 import type { ToolRegistry } from '../tool/framework.js'
 import type { ToolExecutor } from '../tool/executor.js'
 
@@ -94,6 +96,8 @@ export interface RunnerOptions {
   readonly loopDetection?: LoopDetectionConfig
   /** Maximum cumulative tokens (input + output) allowed for this run. */
   readonly maxTokenBudget?: number
+  /** Optional context compression strategy for long multi-turn runs. */
+  readonly contextStrategy?: ContextStrategy
 }
 
 /**
@@ -191,6 +195,10 @@ const ZERO_USAGE: TokenUsage = { input_tokens: 0, output_tokens: 0 }
  */
 export class AgentRunner {
   private readonly maxTurns: number
+  private summarizeCache: {
+    oldSignature: string
+    summaryMessage: LLMMessage
+  } | null = null
 
   constructor(
     private readonly adapter: LLMAdapter,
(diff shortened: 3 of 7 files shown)


JackChen-me:
Three issues found, two are blockers:

**1. `summarize` produces orphaned `tool_result_id` after the 2nd compaction.** At `runner.ts:277`, `splitAt = Math.max(2, Math.floor(rest.length / 2))` can cut through a `tool_use`/`tool_result` pair once `rest` is odd-length (happens after the first summary injects its own user turn). Anthropic hard-rejects the next call with `tool_result block missing corresponding tool_use block`. Fix: align `splitAt` to an even offset, e.g. `Math.max(2, Math.floor(rest.length / 4) * 2)`.

**2. Summary `adapter.chat()` usage isn't added to `totalUsage`.** At `runner.ts:311`, the summary response's usage is emitted in a trace event but never fed back into the accumulator, so `maxTokenBudget` is silently bypassed and `RunResult.tokenUsage` under-reports cost. Fix: return `{ messages, usage }` from `summarizeMessages` and accumulate at the call site.

**3. Synthetic marker/summary messages use `role: 'user'`**, producing sequences like `[user, user, assistant, user]`. Not fatal on Anthropic direct API (silent merge per [anthropic-sdk-typescript#565](https://github.com/anthropics/anthropic-sdk-typescript/issues/565)), but breaks Bedrock and blurs prompt semantics on direct API (summary text concatenates onto the original prompt). Fix: tag as `role: 'assistant'` or fold into the next real user message.

Repros on request.

ibrahimkzmv:
Appreciate the review, you caught a subtle edge case with the odd-length rest 

Fixed a few things:

- safe boundary split for tool_result alignment
- usage tracking now includes summarizeMessages cost
- removed synthetic user rows, moved notes into next turn

Thanks again for the sharp eye @JackChen-me

PR #122, opened 2026-04-17
feat: add dag html

ibrahimkzmv:
## What
This PR introduces a web-based dashboard for visualizing multi-agent task execution as a directed acyclic graph (DAG)
After each run, a static HTML file is generated in the `oma-dashboards` folder

## Why
Multi-agent workflows quickly become difficult to reason about without visibility into execution flow and state.

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes


JackChen-me (changes requested):
Walked through the diff. DAG layout + pan/zoom + detail panel all work as drawn in #4. A few things before merge.

Scope I left fuzzy in #4, my bad:
- The library shouldn't touch the filesystem. `runTeam` writing to `process.cwd()/oma-dashboards/` makes `OpenMultiAgent` unusable in Lambda, browser, or read-only FS. Better split: expose a pure `renderTeamRunDashboard(result: TeamRunResult): string` from `src/dashboard/`, and have the CLI call it and write the file. Keeps the framework free of IO, makes the render function trivially unit-testable, and lets library consumers send the HTML wherever they want (disk, S3, stdout, HTTP response). I'll update #4 with the interface sketch.
- "No new deps" was ambiguous. HTML relies on Tailwind play CDN + Google Fonts + Material Symbols at runtime, none of which work offline. Document the network requirement in `docs/cli.md` for this PR; a follow-up can inline or bundle them.
- `oma-dashboards/` should be in `.gitignore` (CLI-side, once the split above lands).

Security (real blockers):
- Inline JSON in a `<script>` tag is unsafe because `JSON.stringify` doesn't escape `</script>`. Fix: move the data into `<script type="application/json" id="oma-data">...</script>`, read via `JSON.parse(document.getElementById('oma-data').textContent)`, and still escape `</script>` in the payload (e.g. `.replace(/<\/script/gi, '<\\/script')`). The HTML parser still terminates on `</script>` regardless of the script type, so escaping is not optional. But because `application/json` content is never executed, the worst case degrades from arbitrary code execution to a broken DOM or `JSON.parse` error.
- The node render interpolates `${task.title}` into `innerHTML` via a template literal, which injects task titles as raw HTML. Task titles come from the coordinator LLM or `runTasks` user input. Rewrite the node render to use `document.createElement` + `textContent` throughout instead of template literals + `innerHTML`. The render loop already mixes both styles, worth going fully structural.

Tests: checklist has "Added/updated tests" unchecked. Could you add:
- Unit test for `layoutTasks` (pure function, topological order is the interesting case).
- Security regression test for `renderTeamRunDashboard`: feed a task with title `"</script>"` and assert the output contains no unescaped `</script>` and no executable HTML attribute. This is what the two security fixes above exist to prevent, so it should stay locked in.

Architecture:
- `orchestrator.ts` went from ~700 to ~1720 lines, mostly an HTML string. The render function extraction above handles this naturally.
- `emitRunTeamDashboard(tasks: Task[])`: the short-circuit branch fabricates a synthetic `Task` with `description`, `result`, `createdAt`, `updatedAt` that the dashboard never reads, just to satisfy the type. Signature should be `DashboardTaskNode[]` with mapping at each callsite (also moot after the render function extraction).

Nits:
- `buildRunTeamDashboardHtml(_goal, _tasks)`: underscore prefix implies unused, both are used.
- Material Symbols `<link>` is duplicated in head.
- Trailing `<script>` block re-queries `.node` and attaches a second click handler. The render loop already attaches handlers with the `task` closure; trailing one only does `panel.classList.remove("hidden")` and misses re-rendered nodes. Can drop it and fold `closePanel` + outside-click into the main script.
- Stray blank lines around new `scStartMs`/`scEndMs`.

ibrahimkzmv:
I've made the changes. Could you please take another look when you are available @JackChen-me ?

src/dashboard/render-team-run-dashboard.ts
```diff
@@ -0,0 +1,516 @@
...
+            pending: { border: "border-outline", icon: "hourglass_empty", iconColor: "text-outline", container: "bg-surface-container-low opacity-60 grayscale", statusColor: "text-on-surface-variant", chip: "WAITING" },
+        };
+
+        function durationText(task) {
+            const ms = task?.metrics?.durationMs ?? 0;
+            const seconds = Math.max(0, ms / 1000).toFixed(1);
+            return task.status === "completed" ? "DONE (" + seconds + "s)" : task.status.toUpperCase();
+        }
+
+        function layoutTasks(taskList) {
```
JackChen-me: Near-verbatim re-transcription of `src/dashboard/layout-tasks.ts:22-92`: same algorithm, same magic numbers (`nodeW=256`, `colGap=96`, `padX=120`, etc.), but only the server copy has tests. Any tweak on one side silently diverges. Cleanest fix: call `layoutTasks` on the server during render, serialize `positions` into the JSON payload above, let the browser read positions instead of recomputing.

src/orchestrator/orchestrator.ts
```diff
@@ -702,6 +715,14 @@ export class OpenMultiAgent {
   private readonly teams: Map<string, Team> = new Map()
   private completedTaskCount = 0
 
+  private emitRunTeamDashboard(goal: string, nodes: readonly DashboardTaskNode[]): void {
```
JackChen-me: Smuggling the HTML through `{ kind: 'runTeam_dashboard', html }` bypasses the typed `OrchestratorEvent` union; the CLI at `src/cli/oma.ts:333` sniffs for this shape with untyped casts. Once `TeamRunResult` gains `goal`/`tasks` per the spec, the CLI can call `renderTeamRunDashboard(result)` directly after the run returns and the whole event channel can go away. Also fixes the `process.exit` race from blocker 2.

tests/dashboard-render.test.ts
```diff
@@ -0,0 +1,35 @@
+import { describe, expect, it } from 'vitest'
+import { renderTeamRunDashboard } from '../src/dashboard/render-team-run-dashboard.js'
+
+describe('renderTeamRunDashboard', () => {
+  it('does not embed unescaped script terminators in the JSON payload and keeps XSS payloads out of HTML markup', () => {
```
JackChen-me, line 5: Round 1 ask was title + description + result. This only covers `title`. If the renderer ever changes how it embeds description or result, the test won't catch it. Two more cases would lock the invariant.

tests/dashboard-layout-tasks.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  it('orders independent roots in the same column with distinct rows', () => {
+    const tasks = [
+      { id: 'a', dependsOn: [] as const },
+      { id: 'b', dependsOn: [] as const },
+    ]
+    const { positions } = layoutTasks(tasks)
+    expect(positions.get('a')!.x).toBe(positions.get('b')!.x)
+    expect(positions.get('a')!.y).not.toBe(positions.get('b')!.y)
+  })
```
JackChen-me, line 37: Linear, diamond, and multiple-roots covered, but no cycle case. A cyclic input to `layoutTasks` currently silent-loops (the `indegree` counter never hits zero). Add a throw-or-degrade assertion.

src/index.ts
```diff
@@ -58,6 +58,10 @@ export { OpenMultiAgent, executeWithRetry, computeRetryDelay } from './orchestra
 export { Scheduler } from './orchestrator/scheduler.js'
 export type { SchedulingStrategy } from './orchestrator/scheduler.js'
 
+export { renderTeamRunDashboard, escapeJsonForHtmlScript } from './dashboard/render-team-run-dashboard.js'
```
JackChen-me: Nit: `escapeJsonForHtmlScript`, `layoutTasks`, `LayoutTaskInput`, `LayoutTasksResult` aren't framework-consumer APIs. Keep them internal to `src/dashboard/`.

src/cli/oma.ts
```diff
@@ -312,6 +313,44 @@ function mergeOrchestrator(base: OrchestratorConfig, ...partials: OrchestratorCo
   return o
 }
 
+async function writeRunTeamDashboardFile(html: string): Promise<string> {
+  const directory = join(process.cwd(), 'oma-dashboards')
+  await mkdir(directory, { recursive: true })
+  const stamp = new Date().toISOString().replaceAll(':', '-')
```
JackChen-me: Nit: `...replaceAll(':','-')` leaves the `.` before millis (`2026-04-19T12-00-00.123Z.html`). Cosmetic.

JackChen-me (changes requested):
XSS fixes and HTML extraction look clean, thanks for the round 1 pass. Two blockers below, smaller items left as inline comments. 562 tests pass, `tsc --noEmit` clean.

### Blockers

1. **Align with the #4 spec comment.** Round 1 review pointed to #4 for the framework/CLI split sketch, and the renderer signature was also stated directly in round 1: `renderTeamRunDashboard(result: TeamRunResult): string`. `src/dashboard/render-team-run-dashboard.ts:15` still takes a bespoke `TeamRunDashboardInput`, and `TeamRunResult` didn't gain the `goal`/`tasks` fields. `--dashboard` flag (default off, currently missing) and the domain-neutral type names (`TaskExecutionMetrics`/`TaskExecutionRecord`, replacing the `DashboardTaskNode` I named in round 1, so trace exporters can reuse them later) are in the #4 comment.

2. **Dashboard write races `process.exit`, can truncate.** This one's mine, missed it last round. `src/cli/oma.ts:343` fires `void writeRunTeamDashboardFile(html).catch(...)` from `onProgress`; nothing awaits it. `emitRunTeamDashboard` fires inside `runTeam` at three points: short-circuit (`src/orchestrator/orchestrator.ts:950`), budget-exceeded (`:1019`), post-queue (`:1075`). On the first two paths there's no synthesis LLM call between emit and `runTeam` resolving, so the window before `process.exit` is milliseconds. Slow disk, first-time `mkdir`, or AV scan will leave a zero-byte HTML while the exit code still claims success. Two ways to fix: buffer the HTML in the CLI and `await writeRunTeamDashboardFile(html)` after `runTeam` resolves, or collapse the event channel into a direct `renderTeamRunDashboard(result)` call once `TeamRunResult` gains `goal`/`tasks` (fixes the event-channel comment too).

Once 1-2 land the merge is unblocked.

ibrahimkzmv:
I’ve incorporated your feedback and pushed the updates. Would appreciate another review

JackChen-me (approved):
Both blockers resolved: renderer is now a pure `renderTeamRunDashboard(result: TeamRunResult): string` in `src/dashboard/`, `TeamRunResult` carries `goal` + domain-neutral `TaskExecutionRecord[]`, CLI gates writes behind `--dashboard` and awaits the write before exit. XSS mitigations in place (`application/json` script tag, `</script>` escape, `textContent` node render). 565 tests pass, typecheck clean. Merging — I'll push a small follow-up to the `docs/cli.md` wording. Thanks for the thorough iteration.

PR #145, opened 2026-04-22
feat: add translation/backtranslation cookbook example

### LINKED ISSUE #132: Example: translation + backtranslation quality check (cross-model)
## Context

Single-model translation self-check suffers from homogeneity bias: the model thinks its own translation is fine. Using different model families for translate vs. backtranslate removes this. One of the strongest arguments for multi-model orchestration, but no example demonstrates it yet.

## Task

Add `examples/cookbook/translation-backtranslation.ts` (new `cookbook/` directory for use-case-driven examples; create if missing):

1. Agent A: translate EN to target language using Claude
2. Agent B: back-translate to EN using a different provider (OpenAI or Gemini)
3. Agent C: compare original vs. back-translation, flag semantic drift per paragraph

Output: a structured table of `{ original, translation, backtranslation, drift_severity, notes }`.

## Reference

- `examples/basics/multi-model-team.ts`: cross-provider setup
- `examples/patterns/structured-output.ts`: Zod-typed final report

## Acceptance

- Works on a 3-5 paragraph sample of technical English included inline
- Translate and backtranslate use different provider families (Anthropic + OpenAI, or Anthropic + Gemini). Same-family models are not allowed
- Final output validates against a fixed Zod schema: `{ original, translation, backtranslation, drift_severity: 'none'|'minor'|'major', notes }`
- If only one provider's API key is set, the example exits 0 with a readable skip message

## Notes

Three agents, two+ provider families. This is the hardest multi-model use case to replicate with a single-model framework.

zouhh22333-beep:
Closes #132

## What
This PR adds `examples/cookbook/translation-backtranslation.ts` as a cross-model translation QA example.

It implements a 3-agent pipeline:
1. Agent A translates technical English into Simplified Chinese with Claude
2. Agent B back-translates into English using a different provider family
3. Agent C compares original vs. backtranslation and flags semantic drift per paragraph

## Acceptance checklist
- [x] Works on an inline 4-paragraph sample of technical English
- [x] Translation and backtranslation use different provider families
- [x] Final output validates against a fixed Zod schema with:
  - `original`
  - `translation`
  - `backtranslation`
  - `drift_severity` (`none | minor | major`)
  - `notes`
- [x] If the required provider keys are not available, the example exits `0` with a readable skip message

## Notes
For my local run, Agent B uses an OpenAI-compatible endpoint backed by Gemini.

So while the underlying model is Gemini, the framework configuration is set up as an `openai` provider rather than the native `gemini` provider.

I wanted to call this out explicitly in case you prefer the example to use the native Gemini provider path instead.

## Local validation
- ✅ example runs end-to-end successfully
- ✅ `npm run lint` passes
- ⚠️ `npm test` does not fully pass in my local Windows cmd environment due to existing bash-related tests that expect `/bin/bash` / WSL-like behavior; this appears unrelated to this examplethis example

### DIFF
--- a/examples/cookbook/translation-backtranslation.ts
+++ b/examples/cookbook/translation-backtranslation.ts
@@ -1,0 +1,327 @@
+/**
+ * Translation + Backtranslation Quality Check (Cross-Model)
+ *
+ * Demonstrates:
+ * - Agent A: translate EN -> target language with Claude
+ * - Agent B: back-translate -> EN with a different provider family
+ * - Agent C: compare original vs. backtranslation and flag semantic drift
+ * - Structured output with Zod schemas
+ *
+ * Run:
+ *   npx tsx examples/cookbook/translation-backtranslation.ts
+ *
+ * Prerequisites:
+ *   ANTHROPIC_API_KEY must be set
+ *   and at least one of OPENAI_API_KEY / GEMINI_API_KEY must be set
+ */
+
+import { z } from 'zod'
+import {
+  Agent,
+  AgentPool,
+  ToolRegistry,
+  ToolExecutor,
+  registerBuiltInTools,
+} from '../../src/index.js'
+import type { AgentConfig } from '../../src/types.js'
+
+// ---------------------------------------------------------------------------
+// Inline sample text (3-5 technical paragraphs, per issue requirement)
+// ---------------------------------------------------------------------------
+
+const SAMPLE_TEXT = `
+Modern CI/CD pipelines rely on deterministic builds and reproducible environments.
+A deployment may fail even when the application code is correct if the runtime,
+dependency graph, or container image differs from what engineers tested locally.
+
+Observability should combine logs, metrics, and traces rather than treating them
+as separate debugging tools. Metrics show that something is wrong, logs provide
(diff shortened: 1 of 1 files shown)


JackChen-me:
Yeah, let's go with the native gemini path as you flagged.

The openai branch's default `'gemini-2.5-pro'` fails against `api.openai.com` for any user with a plain OpenAI key (unknown model). For the example default, let the openai branch run a real OpenAI model (e.g. `gpt-5.4`, matching `examples/basics/multi-model-team.ts`) and prefer `provider: 'gemini'` when `GEMINI_API_KEY` is set. Pointing the openai SDK at a Gemini-compat endpoint via `baseURL` still works, just not as the default here.

zouhh22333-beep:
Thanks — updated.

The example now prefers the native Gemini provider when `GEMINI_API_KEY` is set, and falls back to `openai` otherwise. The OpenAI default model is now `gpt-5.4`, while `OPENAI_MODEL` + `OPENAI_BASE_URL` can still be used for local testing against OpenAI-compatible endpoints.

JackChen-me:
Merged, thanks for the clean second iteration.

PR #149, opened 2026-04-22
feat: enforce optional outputSchema validation and wire defineTool passthrough (#120)

### LINKED ISSUE #120: [P2] Optional output schema validation for tools
## Motivation

Tool inputs get Zod validation. Tool outputs don't — they're raw strings passed straight to the LLM.

A tool that returns truncated JSON (e.g. API timeout mid-response) produces output that is structurally valid but semantically broken. The framework treats it as a successful result. The LLM hallucinates the missing context and downstream tasks in the DAG inherit the hallucination via SharedMemory.

**Source:** crewAIInc/crewAI#5468 — user reported burning 200k tokens before catching a loop caused by exactly this failure mode. We audited our own pipeline and confirmed the same gap exists in `executor.ts`.

**Priority:** P2 — real production risk but no user has reported it in our repo yet. Loop Detection + Token Budget provide partial mitigation (cap the damage, don't prevent it). Fix is small and contained.

## Proposal

Add an optional `outputSchema` field to `defineTool()`:

```typescript
defineTool({
  name: 'fetch_data',
  schema: z.object({ url: z.string() }),
  outputSchema: z.object({
    data: z.array(z.object({ id: z.string() })),
    complete: z.boolean()
  }),
  execute: async ({ url }) => { ... }
})
```

Behavior:
- `outputSchema` is optional. No schema = current behavior (raw string passthrough)
- When set, `executor.ts` runs `outputSchema.safeParse()` on the tool result
- Validation failure → `isError: true` result with Zod issue details, same as input validation
- No new dependencies (Zod is already a runtime dep)

## Scope

- `src/tool/executor.ts` — add output validation after `tool.execute()`
- `src/tool/define.ts` — accept `outputSchema` in tool definition
- `src/types.ts` — add field to `ToolDef`
- Tests for: valid output passes, truncated JSON fails, missing schema = no-op

Xin-Mai:
## What

This PR adds optional tool output validation using Zod, so each tool can validate its ToolResult.data before the result is returned to the agent loop.

- **Core feature**

  - Added optional outputSchema support for tools and wired it through defineTool/framework registration.
  - Enforced runtime validation of ToolResult.data in the executor when a tool-level schema is configured.
  - Added test coverage for output-schema behavior.
**Primary files**: executor.ts, framework.ts, types.ts, tool-executor.test.ts

- **Follow-up docs/examples pass**
  - Updated English/Chinese docs to reflect the feature and usage.
  - Included stale-example cleanup in README docs to keep field naming current.

**Primary files**: README.md, README_zh.md

- **Scope notes**
In scope: tool-level output-schema feature, executor validation behavior, regression safety, type alignment, and related docs.

Out-of-scope but included as cleanup: stale README example naming updates, called out separately to keep review scope clear.

## Why

Issue [#120 ](https://github.com/JackChen-me/open-multi-agent/issues/120) requests stronger guardrails against malformed tool outputs (especially JSON-like text that looks plausible but is invalid/incomplete). This change catches those failures at tool execution time and surfaces clear errors instead of letting invalid output propagate.

Fixes [#120 ](https://github.com/JackChen-me/open-multi-agent/issues/120)

## Backward compatibility
Existing tools without outputSchema are unaffected.
ToolResult.data remains string-based.
Type defaults preserve current usage patterns.
## Validation performed
npm run lint passes.
npm test -- tests/tool-executor.test.ts passes.
npm test -- tests/built-in-tools.test.ts passes.
Full test suite passes locally.

## Checklist

- [✅] `npm run lint` passes
- [✅] `npm test` passes
- [✅] Added/updated tests for changed behavior
- [✅] No new runtime dependencies (or justified in the PR description)

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,15 @@ Long tool outputs can blow up conversation size and cost. Two controls work toge
 
 **Validation (optional).** Add `outputSchema` to catch malformed tool results before they are forwarded:
 
+> **Note — two different `outputSchema` fields.** The one on `defineTool()` /
+> `ToolDefinition` (shown below) validates a single **tool's** `ToolResult.data`
+> — it is always a `ZodSchema<string>` because tool output is serialised as
+> text. The `outputSchema` on [`AgentConfig`](examples/patterns/structured-output.ts)
+> is different: it validates the **agent's final answer** as parsed JSON
+> against an arbitrary Zod schema (see _Structured output_ in `examples/`).
+> Different types, different scopes — TypeScript won't warn you if you mix
+> them up, so pick the one that matches the layer you're working at.
+
 ```typescript
 const jsonTool = defineTool({
   name: 'json_tool',
--- a/README.md
+++ b/README.md
@@ -267,8 +267,8 @@ import { z } from 'zod'
 const weatherTool = defineTool({
   name: 'get_weather',
   description: 'Look up current weather for a city.',
-  schema: z.object({ city: z.string() }),
-  execute: async ({ city }) => ({ content: await fetchWeather(city) }),
+  inputSchema: z.object({ city: z.string() }),
+  execute: async ({ city }) => ({ data: await fetchWeather(city) }),
 })
 
 const agent: AgentConfig = {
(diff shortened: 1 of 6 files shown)


JackChen-me (commented):
Main thing: `outputSchema` validates error results too. If a tool returns `{ data: 'API timeout', isError: true }` and the schema expects JSON, the real error gets overwritten by a Zod issue string. `bash`, `file_read`, `grep` all produce `isError: true` in practice. Gate on `!result.isError` in executor.ts and add a test.

A few smaller things. `outputSchema` name collides with `AgentConfig.outputSchema`, which validates final agent output. Different types so no compile conflict, but worth a README note to disambiguate. `TOutput = string` default is misleading since runtime always validates the string `data`; passing `z.object({...})` type-checks but always fails at runtime. Consider narrowing to `ZodSchema<string>` or a doc note. Three stray semicolons in executor.ts at lines 150/168/199, rest of file is no-semi. The `schema:/content:` README fixes address a stale example; worth calling out separately in the PR body so the scope is clear.

Will merge once the isError gate is added.

Xin-Mai:
Thanks a lot for your review, I will fix them

Issue #136, opened 2026-04-20
Example: contract review pipeline with task DAG


JackChen-me:
## Context

Document processing pipelines need step-level retry, not whole-pipeline retry. OMA's task DAG + retry fits this cleanly. Good example to show DAG dependencies and failure isolation together.

## Task

Add `examples/cookbook/contract-review-dag.ts` (new `cookbook/` directory for use-case-driven examples; create if missing):

- Task 1: extract clauses (input: contract text, output: structured clause list)
- Task 2: compliance check (depends on Task 1, flags risky clauses)
- Task 3: summary (depends on Task 1, generates exec summary)
- Task 4: notify (depends on Task 2 + 3, outputs final Markdown)

Each task retries on failure. Log retry attempts via `onProgress` (task_retry events).

## Reference

- `examples/basics/task-pipeline.ts`: task DAG
- `examples/patterns/task-retry.ts`: retry config

## Acceptance

- Uses a sample contract as inline text or a short `examples/fixtures/sample-contract.txt`. No PDF parsing
- Task 2 and 3 run in parallel after Task 1 completes. Assertion: their start times are within 500ms of each other
- README includes a forced-failure mode (env `FORCE_FAIL=task2`) that deterministically fails Task 2's first attempt, proving retry fires at least once
- `onProgress` logs every `task_retry` event with attempt number and task id

## Notes

No PDF parsing libraries. This example is about pipeline shape + retry + DAG, not PDF tooling. Input is clean text.

fault-segment:
Can you assign this to me?I'd like to work on this.

JackChen-me:
@fault-segment  of course

JackChen-me:
Shipped in #155 (thanks @fault-segment). Closing.

PR #155, opened 2026-04-23
feat: add contract-review-dag cookbook example with step-level retry

fault-segment:
## What

This PR add  an contract review pipeline with task DAG example.Parallel execution and forced retries were verified.

## Why
Complete this Issue #136 

## Checklist

- [✅] `npm run lint` passes
- [✅ ] `npm test` passes
- [✅] Added/updated tests for changed behavior
- [✅] No new runtime dependencies (or justified in the PR description)

## Example Result:
截屏2026-04-23 11 28 40


JackChen-me (changes requested):
Traced the FORCE_FAIL path, works (beforeRun throw → Agent returns success:false → executeWithRetry retries).

Main thing: the README update from issue #136 AC is missing. Add an entry in `examples/README.md` matching the style of the other two cookbook entries, mention both run modes.

Two cleanups: `setTimeout(resolve, 100)` before `verifyParallelism` is unnecessary since onProgress fires synchronously. The "Read the clause list from shared memory" phrasing in task descriptions is misleading too; default memoryScope is `'dependencies'`, so dep results are injected into the prompt and the agent isn't reading memory as an action. Reword to "Using the clause list from Task 1 above" or similar.

fault-segment:
thx for detailed review.I've fix them,add the example in the README.md,remove the setTimeout and fix the task desc.

JackChen-me:
LGTM, hits #136.

One ask: the parallelism check logs `YES/NO` but doesn't fail. Issue asked for an assertion, so it should exit nonzero when the diff is >= 500ms (see `meeting-summarizer.ts:213-220` for the pattern). Otherwise a scheduler regression passes silently. Also missing trailing newlines on both new files.

fault-segment:
Fixed that.Now it's anassertion and newlines added.

JackChen-me (approved):
All feedback addressed — README entry, setTimeout removed, task descriptions reworded, parallelism assertion added, trailing newlines. Thanks!

PR #160, opened 2026-04-23
examples: add Engram integration (memory store, toolkit, two demos)

Agentscreator:
Closes #157. 

Adds Engram integration examples using the TeamConfig.sharedMemoryStore and SupportedProvider updates from #157/#158.

Added

1. engram-store.ts — MemoryStore backed by the Engram REST API. Plugs into any team via sharedMemoryStore; also a reference template for custom backends.
2. engram-toolkit.ts — registers engram_commit, engram_query, engram_conflicts, and engram_resolve via ToolRegistry.
3. research-team.ts — manual pattern: agents coordinate explicitly through Engram tools.
4. team-research.ts — automatic pattern: same pipeline via runTeam() with EngramMemoryStore; context flows through Engram without manual tool calls.

Both demos are intentionally included to illustrate the tradeoff between the two approaches.
Nits from #157

SupportedProvider imported from index.js, no local alias.
Removed as AgentProvider cast (no longer needed after #158).

Prerequisites
Requires a running Engram server and workspace invite key — see [engram-memory.com](http://engram-memory.com/).

1. http://localhost:7474
2. ENGRAM_INVITE_KEY env var
3. AGENT_PROVIDER / AGENT_MODEL for the provider API key

### DIFF
--- a/examples/integrations/with-engram/engram-store.ts
+++ b/examples/integrations/with-engram/engram-store.ts
@@ -21,6 +21,7 @@ import type { MemoryEntry, MemoryStore } from '../../../src/types.js'
 
 interface EngramFact {
   fact_id: string
+  lineage_id: string
   content: string
   scope: string
   agent_id?: string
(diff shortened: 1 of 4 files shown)


JackChen-me:
`team-research.ts` doesn't actually wire any Engram tools to the orchestrated agents. The registry at lines 84-86 never reaches `runTeam()`, which builds its own per-agent registry via `buildPool → buildAgent` (`src/orchestrator/orchestrator.ts:1556,216`) from built-ins plus `AgentConfig.customTools`. The `tools: [...engramToolNames]` allowlist at 110/128/145 names tools that aren't registered, so `AgentRunner` (`src/agent/runner.ts:465`) silently filters it to zero. Demo runs, but the `engram_*` calls the prompts demand never happen (LLM isn't told those tools exist). Only the `sharedMemoryStore` auto-plumbing path still works. Fix: add `getTools()` to `EngramToolkit` and pass `customTools: toolkit.getTools()` on each config, drop the outer `registry`.

Two API-behavior checks: `engram-store.ts` `get()` does `limit=1` with no `order_by` (assumes newest-first), `delete()` posts empty content with no confidence. Server accepts both?

`research-team.ts` looks fine.

JackChen-me:
LGTM, merging.

PR #161, opened 2026-04-23
fix: resolve context compaction persistence and turn dropping

### LINKED ISSUE #152: [Bug] Context Compaction Strategy Fails to Persist Across Turns
## Description
The `contextStrategy` implementation inside `AgentRunner.stream()` currently compacts the conversation history internally but fails to return the compacted history back to the caller. The `RunResult` currently only returns `messages` (the new messages generated during the run), resulting in the caller retaining the original, uncompacted history payload and causing potential OOM/Token Budget exhaustion.

Additionally, the context compaction logic is currently skipped on the initial turn (`turns > 1`), meaning an initial prompt that exceeds the maximum token limit will not trigger the compaction strategy and will instead crash or trigger a `TokenBudgetExceededError`.

## Reproduction Steps
1. Configure an `AgentRunner` with a `contextStrategy` (e.g., `'summarize'`) that triggers at 50,000 tokens.
2. Initialize an agent run using a `messageHistory` array containing > 50,000 tokens.
3. Observe that on the initial turn (`turns === 1`), the history is not compacted, triggering an immediate budget error or context overflow.
4. Bypass the first-turn restriction by forcing a multi-turn run. Observe that `AgentRunner` successfully compacts the history internally and generates a summarized prompt.
5. Review the resulting `RunResult.messages`. The `messages` array only contains the new turns, and the compacted history is discarded.
6. The caller appends the new `RunResult.messages` to its ongoing `history` array, meaning the old context was never actually removed.

## Expected Behavior
1. The `contextStrategy` should evaluate the context size on the *first* turn before attempting to execute the LLM call.
2. The `RunResult` object must include the `finalContext` (the fully compacted message history) so the orchestrator/caller can replace its persistent array with the lightweight version.

## Proposed Solution
1. Remove `turns > 1` from the `this.options.contextStrategy` evaluation in `AgentRunner.stream()`.
2. Expand `RunResult` (and `AgentRunResult`) in `src/types.ts` to include an optional `finalContext?: LLMMessage[]` array.
3. Populate `finalContext` at the end of the `stream()` loop and pass it up through the stream events.
4. Downstream orchestrators (like `apollo_cli.ts` or `agent.ts`) can then check for `result.finalContext`. If present, they replace their history entirely rather than blindly pushing new messages.

apollo-mg:
Fixes #152 as discussed. Removed the turns > 1 gate and added a newMessages array in stream() to prevent the slice method from silently dropping turns during summarize/custom compaction strategies.

### DIFF
--- a/src/agent/agent.ts
+++ b/src/agent/agent.ts
@@ -206,7 +206,7 @@ export class Agent {
 
     const result = await this.executeRun([...this.messageHistory])
 
-    // Persist the new messages into history so the next `prompt` sees them.
+    // Persist the new messages into history so the next prompt sees them.
     for (const msg of result.messages) {
       this.messageHistory.push(msg)
     }
--- a/src/agent/runner.ts
+++ b/src/agent/runner.ts
@@ -374,10 +374,12 @@ export class AgentRunner {
       : '[Conversation summary unavailable]'
 
     this.summarizeCache = { oldSignature, summaryPrefix }
+    
     const mergedRecent = prependSyntheticPrefixToFirstUser(
       recentPortion,
       `${summaryPrefix}\n\n`,
     )
+    
     return {
       messages: [firstUser, ...mergedRecent],
       usage: summaryResponse.usage,
@@ -538,6 +540,7 @@ export class AgentRunner {
   ): AsyncGenerator<StreamEvent> {
     // Working copy of the conversation — mutated as turns progress.
     let conversationMessages: LLMMessage[] = [...initialMessages]
+    const newMessages: LLMMessage[] = []
 
     // Accumulated state across all turns.
     let totalUsage: TokenUsage = ZERO_USAGE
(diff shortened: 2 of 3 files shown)


JackChen-me:
Core fix direction matches what I laid out in the issue comment, good.

Few things before merging:

1. The sampling params (topP/topK/minP/freq/presence/extraBody) are dead code. Neither `anthropic.ts` nor `openai.ts` reads them, and `AgentConfig` doesn't expose them either. Also unrelated to #152. Please pull them out of this PR. A separate PR wiring `AgentConfig` → `RunnerOptions` → both adapters end-to-end would be welcome.

2. The actual bug (slice() dropping turns when summarize/custom shrinks `conversationMessages` below `initialMessages.length`) has no test. Bumping `toHaveBeenCalledTimes(2)` only covers the gate removal. Please add a test where `custom.compress` returns fewer messages than the seed and assert `result.messages` still contains the new assistant + tool_result turns.

3. Restore the `// Persist the new messages into history so the next prompt sees them.` comment in `agent.ts`, and drop the two blank lines added around `mergedRecent` in `runner.ts`.

4. Dropping `turns > 1` means `custom.compress` now runs on the initial prompt too. summarize/compact/sliding-window self-gate on size thresholds; custom doesn't. Worth a note on the `contextStrategy.custom` type doc so existing custom strategies aren't surprised.

JackChen-me:
LGTM. Test actually hits the slice bug now. Squash-merging; small cosmetic follow-up on main.

PR #163, opened 2026-04-23
feat: expose advanced LLM sampling parameters

apollo-mg:
**What**

Exposes advanced LLM sampling parameters (topP, topK, minP, frequencyPenalty, presencePenalty, and extraBody) through the public AgentConfig and RunTeamOptions interfaces, passing them down cleanly to the underlying openai and anthropic adapters.

**Why**

While cloud API users (OpenAI/Anthropic) often rely on default sampling parameters, granular control is strictly required for developers running local, highly quantized MoE models (e.g., Qwen, Gemma) on consumer hardware. Without the ability to tune parameters like frequency_penalty or top_k, these models frequently fall into repetitive hallucination loops when strictly adhering to complex, multi-turn tool-calling schemas.

This PR properly wires up the parameters Jack previously identified as "dead code" directly through the RunnerOptions into the adapter HTTP payloads.

**Checklist**

- [x] npm run lint passes
- [x] npm test passes


JackChen-me:
Wiring looks clean, that part matches what I had in mind for #161. A few things to pull out before merge though:

1. `frequency_penalty ?? 0.1` in openai.ts. OpenAI defaults to 0, so this changes output for anyone who never set the field. Pass through undefined.

2. `parallel_tool_calls: ... ? false : undefined`. OpenAI defaults to true. Forcing false makes every multi-tool interaction a serial turn loop (one tool per API round trip instead of batched), which is a latency/token hit for everyone. If this is for your local setup, worth a follow-up PR with an opt-in AgentConfig flag.

3. `reasoning_content` → `<think>...</think>` lands in `fullText` which flows into context/history hooks, and it's OpenAI-only right now. Worth its own PR with an opt-in and parity across adapters.

Small nits: tests checkbox is ticked but the diff has no test changes, a flow test (AgentConfig field set → surfaces in LLMChatOptions) covers it. The `as any` casts are fine but worth a one-line comment noting they're for local OpenAI-compatible servers that accept non-standard params like top_k / min_p.

If you're up for reshaping this PR to just the six sampling params + the test, I'd merge that quickly. The other two can come as follow-ups.

JackChen-me:
LGTM on the wiring, taking the cleanup directly so it can ride the next release. Your commits and co-authorship stay on the merge.

Cleanup I'm pushing:

- Trim the `reasoning_content` line from the description since the stream mapping isn't actually in the diff.
- Drop the `delta as any` cast (with `reasoning_content` out of scope, no non-standard fields are read on the delta). Narrow the payload cast to only `top_k` / `min_p` / `extraBody`. The `streamResponse` cast may need to stay if overload resolution can't pick the streaming return type once the payload is widened.
- Reorder the payload spread so `extraBody` falls between sampling params and structural fields (`model` / `messages` / `tools` / `stream`). Right now spread ordering lets `extraBody` accidentally override `model` / `messages`, which is a footgun.
- Add a wire-format assertion to the test so we lock the actual HTTP body shape, not just the `LLMChatOptions` mapping. Same for the `runTeam` coordinator passthrough.
- Add JSDoc to all six new fields with provider-compat notes:
  - `topP`: universal.
  - `topK`: Anthropic + OpenAI-compatible local; cloud OpenAI rejects.
  - `frequencyPenalty` / `presencePenalty`: OpenAI track only; Anthropic adapter ignores.
  - `minP`: OpenAI-compatible local only; cloud OpenAI rejects.
  - `extraBody`: escape hatch ordered to override sampling defaults but not transport fields.

PR #167, opened 2026-04-24
docs: add OpenRouter provider example

kenrogers:
## Summary
- add a runnable OpenRouter provider example using the existing OpenAI-compatible adapter
- document `OPENROUTER_API_KEY` in the quick start provider env vars
- list OpenRouter in the provider examples and supported providers docs
- add CLI guidance for using OpenRouter with `provider: 'openai'` and `baseURL`

## Why
OpenRouter already works through the OpenAI-compatible adapter, and a few pattern examples reference it, but it was not visible in the main provider example path. This gives users a direct copy/paste example without adding runtime dependencies or a new adapter.

## Test plan
- npm run lint
- npm test


JackChen-me:
LGTM. Nice that orchestrator-level defaults also route the coordinator through OpenRouter.

Issue #147, opened 2026-04-22
[Tracker] Propose a vertical example from your own domain


JackChen-me:
> **First contribution?** Proposing from scratch is the highest-leverage
> but hardest path here. If you'd rather warm up with a pre-scoped example
> task first, see [#148](https://github.com/JackChen-me/open-multi-agent/issues/148)
> for a difficulty ladder.

---

## What this is for

We're not handing out a checklist of examples to build. We're asking you to
bring one.

If you're a student or someone switching into Agent work, you have something
most maintainers don't: real context in a specific domain. Medicine, law,
finance, education, HR, e-commerce, manufacturing, research, game dev,
whatever you've spent time on. The best OMA examples won't come from our
imagination. They'll come from pain you've felt in your field, turned into a
multi-agent solution.

## What we want from you

Comment below using this format:

> **Your background**: e.g. "Second-year law student interning at an M&A firm",
> or "3 years in SaaS customer support, now learning Agent"
>
> **The pain point**: e.g. "Associates spend 4-6 hours per deal manually
> comparing clauses across 10+ redlined NDAs"
>
> **Why multi-agent fits**: why this needs orchestration, not a single LLM call
>
> **Proposed design** (3-5 lines): agents, their roles, which OMA primitives
> (`runTeam` / `runTasks` / shared memory / structured output)
>
> **Reusable pattern**: what can someone outside your domain take away?

## How this works

1. You comment with your proposal using the format above
2. We review in the thread. If scope, multi-agent fit, and story hold up, we
   open a dedicated issue for your example and assign it to you
3. You build on a branch. Follow `CONTRIBUTING.md` and look at the existing
   files under `examples/patterns/` and `examples/cookbook/` for style
4. You open a PR against `main`. We review code quality, runnability, and
   the docs block inside the example file
5. On merge, your example lands in `examples/cookbook/` with your name and
   context attached, gets credited in release notes, and can be linked from
   the README

If a proposal isn't quite there, we comment with what would need to change
instead of closing it. Iterate in-thread, then move to a dedicated issue when
it's ready.

## Why this is worth your time beyond the resume line

Most contributors to OSS tutorials follow a spec someone else wrote. That
doesn't differentiate you in an interview. Proposing your own does.

- You define the problem, so "why this matters" has a real answer
- You own the design, so "why X over Y" isn't hand-waved
- Your example lives in the repo with your handle, linked from the README,
  referenced in release notes
[...]

PR #172, opened 2026-04-25
Add learning assistant cookbook example

orangelab-dev:
## Summary

Add a new cookbook example for an education-domain learning assistant.

## Why

This repo already has examples for summarization, contract review, and competitive monitoring, but it does not yet have a simple education-focused vertical example.

This example is meant to show a practical multi-agent workflow for a common learning problem: turning "I want to learn React" into a usable study plan. It is intentionally small and easy to follow, so new users can understand how to structure a real multi-agent pipeline without extra tools or framework changes.

It is not a single LLM call wrapped as an example:
- `roadmap-planner` creates the staged learning roadmap
- `resource-curator` selects beginner-friendly resources
- `practice-coach` consumes both upstream outputs and produces the final 4-week study plan

## What changed

- added `examples/cookbook/learning-assistant.ts`
- updated `examples/README.md`

## Design notes

- uses `runTasks()` instead of `runTeam()` to keep the workflow explicit and easy to inspect
- keeps the example tool-free with `tools: []`
- adds a per-agent timeout so the demo does not wait indefinitely
- adds token usage output so the example has honest cost visibility
- adds no new runtime dependencies

## Run

```bash
npx tsx examples/cookbook/learning-assistant.ts
```

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -57,6 +57,7 @@ End-to-end examples framed around a concrete problem (meeting summarization, tra
 | [`cookbook/meeting-summarizer`](cookbook/meeting-summarizer.ts) | Fan-out post-processing of a transcript into summary, structured action items, and sentiment. |
 | [`cookbook/contract-review-dag`](cookbook/contract-review-dag.ts) | 4-task DAG (extract → compliance-check + summary → notify) with step-level retry. Run normally or with `FORCE_FAIL=task2` to exercise retry. |
 | [`cookbook/competitive-monitoring`](cookbook/competitive-monitoring.ts) | Parallel source monitoring (Twitter/Reddit/News), contradiction detection, and aggregated intelligence reporting. |
+| [`cookbook/learning-assistant`](cookbook/learning-assistant.ts) | Education-domain planning: a multi-agent team builds a React learning roadmap, starter resources, and a week-by-week practice plan. |
 
 ## integrations — external systems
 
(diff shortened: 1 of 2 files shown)


JackChen-me:
The code's clean but this overlaps with `meeting-summarizer` (fan-out) and `contract-review-dag` (DAG with deps). Swapping the domain to "learn React" doesn't add a framework capability readers can't already see in those.

Would land it if it shows something new. Most useful angle: add a Zod `outputSchema` on `roadmap-planner` (e.g. `{ stages: [{ goal, topics, milestone }] }`) so the example demonstrates structured output and the validation retry, then have `practice-coach` consume the parsed object instead of a markdown blob. While you're in there, drop `maxConcurrency: 1` so the two independent tasks actually run in parallel, and take the learner goal from `process.argv[2]` instead of hardcoding React.

JackChen-me:
Actually, vertical examples go through #147 first. Propose there; if it holds up we open a dedicated issue and you build from that. Technical notes above still apply.

orangelab-dev:
Thanks a lot for the detailed feedback — really helpful.

I see your point about the overlap with existing examples, and the
suggestions on structured output and concurrency make a lot of sense. I’ll
take some time to better understand the framework and revisit this from a
more meaningful angle.

Also noted on the proposal process (#147), I’ll follow that path next time.

Appreciate the guidance!

Jack Chen ***@***.***> 于2026年4月25日周六 16:01写道：

Issue #169, opened 2026-04-25
Native streaming support for reasoning_content / `<think>` blocks


JackChen-me:
# Problem

Reasoning-capable local models (DeepSeek R1, Qwen QwQ, others served via llama-server / vLLM) emit a `reasoning_content` channel alongside `content` during streaming. OMA currently surfaces only `content`, so users running these models lose the model's thinking trace — important for both debugging and end-user UX (showing a `<think>…</think>` block in the response).

# Proposed scope (needs design)

Two parts:

1. Extend `StreamEvent` with a `reasoning` event type so adapters can emit reasoning deltas without conflating them with text content.
2. Map OpenAI streaming `delta.reasoning_content` to the new event in `OpenAIAdapter.stream()`.

Open design questions:

- Retain reasoning content in `LLMResponse.content` as a separate block type, or only surface via streaming?
- Anthropic exposes thinking via `thinking` content blocks (Claude Sonnet 3.7+) under a different schema — should one event type cover both, or two distinct?
- Does retaining reasoning across turns count toward `maxTokenBudget`?

# Background

Came up during PR #163 review. The original PR from @apollo-mg included a `<think>`-tag parser in `OpenAIAdapter.stream()`, which we pulled out because it deserves an explicit design pass on `StreamEvent` shape rather than a side-channel hack. Tracking here so the use case isn't lost; not a green light to implement until the StreamEvent extension is designed.

SiMinus:
Hi! bro  I’d be happy to work on this issue. Would you be willing to assign it to me?

JackChen-me:
@SiMinus  of course

JackChen-me:
Shipped in #174 (thanks @SiMinus). Closing.

PR #174, opened 2026-04-25
feat(streaming): add reasoning events for OpenAI and Anthropic

SiMinus:
Implemented native reasoning-streaming support with a minimal, explicit extension of the streaming contract.

## What changed

| Area | Change |
| --- | --- |
| Streaming contract | Added a new `reasoning` event to `StreamEvent`, so reasoning deltas are no longer conflated with normal text. |
| Response model | Added a `ReasoningBlock` content type and retained reasoning in `LLMResponse.content` instead of exposing it only during streaming. |
| OpenAI streaming | Updated `OpenAIAdapter.stream()` to map `delta.reasoning_content` to `reasoning` events and preserve the full reasoning trace in the final `done` response. |
| OpenAI non-streaming | Updated OpenAI completion parsing so non-streaming responses can also retain `reasoning_content`. |
| Anthropic mapping | Mapped Anthropic `thinking` / `thinking_delta` into the same unified `reasoning` event/block shape. |
| Cross-provider serialization | Updated message serialization paths so retained reasoning can safely round-trip through provider adapters. |
| Runner output | Updated the runner text extraction path to render retained reasoning as `<think>...</think>` instead of dropping it. |

## Tests

- ✅ `npm run lint`
- ✅ `npm test -- tests/openai-adapter.test.ts tests/anthropic-adapter.test.ts`
- ✅ `npm test` (`48` test files, `688` tests)

## My solutions to open design questions

- `LLMResponse.content`: I retained reasoning in the final response as a separate `ReasoningBlock`, rather than surfacing it only during streaming.
- Anthropic `thinking`: I mapped Anthropic `thinking` / `thinking_delta` and OpenAI-compatible `reasoning_content` into one unified `reasoning` event/block, not two provider-specific event types.
- `maxTokenBudget`: In this implementation, retained reasoning is carried forward as part of the conversation context, so it effectively counts toward later token budget usage; this is an implementation consequence, not a separately designed budgeting rule.

### DIFF
--- a/src/agent/runner.ts
+++ b/src/agent/runner.ts
@@ -191,14 +191,11 @@ export interface RunResult {
 // Internal helpers
 // ---------------------------------------------------------------------------
 
-/** Extract renderable assistant text, including reasoning blocks as `<think>` tags. */
+/** Extract every TextBlock from a content array and join them. */
 function extractText(content: readonly ContentBlock[]): string {
   return content
-    .flatMap((b): string[] => {
-      if (b.type === 'reasoning') return [`<think>${b.text}</think>`]
-      if (b.type === 'text') return [b.text]
-      return []
-    })
+    .filter((b): b is TextBlock => b.type === 'text')
+    .map(b => b.text)
     .join('')
 }
 
--- a/src/agent/runner.ts
+++ b/src/agent/runner.ts
@@ -191,11 +191,14 @@ export interface RunResult {
 // Internal helpers
 // ---------------------------------------------------------------------------
 
-/** Extract every TextBlock from a content array and join them. */
+/** Extract renderable assistant text, including reasoning blocks as `<think>` tags. */
 function extractText(content: readonly ContentBlock[]): string {
   return content
-    .filter((b): b is TextBlock => b.type === 'text')
-    .map(b => b.text)
+    .flatMap((b): string[] => {
+      if (b.type === 'reasoning') return [`<think>${b.text}</think>`]
+      if (b.type === 'text') return [b.text]
+      return []
+    })
     .join('')
 }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -128,6 +128,7 @@ export { SharedMemory } from './memory/shared.js'
 
 export type {
   // Content blocks
+  ReasoningBlock,
   TextBlock,
   ToolUseBlock,
   ToolResultBlock,
--- a/src/llm/anthropic.ts
+++ b/src/llm/anthropic.ts
@@ -42,6 +42,7 @@ import type {
   LLMMessage,
   LLMResponse,
   LLMStreamOptions,
+  ReasoningBlock,
   LLMToolDef,
   StreamEvent,
   TextBlock,
@@ -62,6 +63,10 @@ import type {
  */
 function toAnthropicContentBlockParam(block: ContentBlock): ContentBlockParam {
   switch (block.type) {
+    case 'reasoning': {
+      const param: TextBlockParam = { type: 'text', text: `<think>${block.text}</think>` }
+      return param
+    }
     case 'text': {
       const param: TextBlockParam = { type: 'text', text: block.text }
       return param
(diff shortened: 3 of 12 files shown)


JackChen-me (changes requested):
Two regressions before this can merge.

1. `runner.ts:195` `extractText` prepends `<think>...</think>` to the final text. With `outputSchema`, that flows into `extractJSON`'s case-3 fallback (`indexOf('{')` to `lastIndexOf('}')`), which grabs the wrong span whenever the reasoning contains `{`. Either skip reasoning blocks in `extractText` or use a text-only extractor for the structured path. No test covers the combo.

2. `fromAnthropicContentBlock` and the `content_block_delta` switch use `(block as { ... }).field` casts. Unnecessary: SDK 0.52 already has `ThinkingBlock` and `ThinkingDelta` in the relevant unions (`messages.d.ts:138, 326`), so `case 'thinking'` / `case 'thinking_delta'` on the original typed switches narrows correctly. The default branch also loses exhaustiveness as written.

Side note: `ThinkingBlock.signature` is dropped on the way in, which breaks extended-thinking continuation. OK to handle as a follow-up.

JackChen-me:
@codex

SiMinus:
Thanks for the review. I addressed both blocking regressions.

Changes made:

| Review item | Fix |
| --- | --- |
| `extractText` prepended `<think>...</think>` into final output | Updated `AgentRunner` text extraction to ignore `ReasoningBlock` and only join `TextBlock`s, so structured-output parsing no longer sees reasoning content. |
| Missing structured-output coverage | Added a regression test where reasoning contains a misleading `{` and the final text contains valid JSON. The structured output path now validates the JSON correctly. |
| Anthropic casts bypassed SDK narrowing | Removed the unnecessary `(block as ...)` / `(delta as ...)` casts and switched directly on `block.type` / `delta.type`, restoring SDK union narrowing and exhaustiveness behavior. |

I left `ThinkingBlock.signature` out of this PR since it was called out as acceptable follow-up work.

Validation:

- ✅ `npm test -- tests/structured-output.test.ts tests/anthropic-adapter.test.ts tests/openai-adapter.test.ts`
- ✅ `npm run lint`
- ✅ `npm test` (`48` test files, `689` tests)

SiMinus:
Implemented the review follow-ups for native reasoning support.

This update keeps the original reasoning streaming design, while fixing the merge-blocking regressions and token-estimation gap:

| Area | Update |
| --- | --- |
| Structured output | `AgentRunner` now excludes `ReasoningBlock` from final text extraction, so reasoning traces cannot interfere with `outputSchema` JSON parsing. |
| Regression coverage | Added a structured-output test where reasoning contains a misleading `{` and final text contains valid JSON. |
| Anthropic adapter | Removed unnecessary casts and restored SDK union narrowing for `thinking` / `thinking_delta`. |
| Token estimation | `estimateTokens()` now counts retained reasoning text so context compaction and summarization thresholds account for reasoning traces. |
| Token test coverage | Added a focused test for reasoning token estimation. |

Validation:

- ✅ `npm run lint`
- ✅ `npm test -- tests/structured-output.test.ts tests/anthropic-adapter.test.ts tests/openai-adapter.test.ts`
- ✅ `npm test -- tests/tokens.test.ts tests/context-strategy.test.ts tests/structured-output.test.ts`
- ✅ `npm test` (`49` test files, `690` tests)

JackChen-me:
Thanks, the latest round addresses the structured-output and SDK-narrowing regressions cleanly. One outbound-serialization concern before merge.

`anthropic.ts:67`, `openai-common.ts:143/179`, and `gemini.ts:88` wrap reasoning text as `<think>...</think>` and push it back into provider history as ordinary assistant/user text. The content was real, but the format is non-roundtrip: what the model emitted (e.g. an Anthropic `ThinkingBlock` plus `TextBlock`) gets flattened into a single text block on the next turn, and the model sees a different shape than what it actually produced.

Concrete costs:

- Context pollution: `<think>...</think>` becomes part of assistant text on subsequent turns, where the model has no contract that says "ignore this region."
- Token waste: reasoning traces are typically the largest part of a turn, and they now ride along on every subsequent request.
- Blocks the future Anthropic native path: extended-thinking continuation across tool-use turns expects `ThinkingBlockParam` (with `signature`) in history, not text. Once we want that, the text-wrap codepath has to come out anyway.

Simplest fix: skip reasoning blocks when serializing outbound. Stream events still fire, `LLMResponse.content` still retains reasoning for consumers, history just stops carrying it.

```ts
// anthropic.ts toAnthropicContentBlockParam
-    case 'reasoning': {
-      const param: TextBlockParam = { type: 'text', text: `<think>${block.text}</think>` }
-      return param
-    }
```

Same delete in `toOpenAIUserMessage`, `toOpenAIAssistantMessage`, and `toGeminiContents` (and the `toThinkTag` helper goes with them). Anthropic native thinking continuation (with signature) is a separate follow-up. Main currently degrades thinking blocks to a fallback text block anyway, so this PR doesn't introduce a regression on that front.

One caveat worth flagging: local `<think>`-tag models (DeepSeek R1, QwQ) emit reasoning as part of their normal text output, not as a separate channel, so they're unaffected by the outbound rule above; their tags will keep appearing in `text` blocks the way they do today. If we want a different policy for those later, it can be a separate issue.

---

Process note for next time, not blocking this PR: #169 explicitly says "not a green light to implement until the StreamEvent extension is designed," and lists three open design questions. This PR ended up answering all three (retain reasoning in `LLMResponse.content`, unify Anthropic + OpenAI into one event/block, count toward token budget). The answers themselves are reasonable, but the agreement should land in the issue before the implementation PR opens, so reviewers can push back on the design separately from the code. I'll split those into their own issues so they're tracked, not buried in this thread:

- Retain reasoning in `LLMResponse.content` vs. stream-only
- Unify or split Anthropic thinking vs. OpenAI `reasoning_content` block types
- Token budget treatment for reasoning text

Drop those four cases and this is good to go.

SiMinus:
Thanks, addressed the outbound serialization concern.

What changed:

| Area | Update |
| --- | --- |
| Anthropic outbound history | `ReasoningBlock` is now filtered out before converting framework messages to Anthropic message content. |
| OpenAI outbound history | Removed the `<think>...</think>` wrapping path from both user and assistant message serialization. |
| Gemini outbound history | `ReasoningBlock` is now filtered out before converting framework messages to Gemini parts. |
| Preserved behavior | `reasoning` stream events still fire, `LLMResponse.content` still retains `ReasoningBlock`, and token estimation still counts retained reasoning. |

This means provider history no longer gets polluted with reasoning text serialized as ordinary assistant/user text. Local models that emit literal `<think>` tags inside normal text blocks are unchanged, as noted.

Validation:

- ✅ `npm run lint`
- ✅ `npm test -- tests/openai-adapter.test.ts tests/anthropic-adapter.test.ts tests/gemini-adapter-contract.test.ts tests/tokens.test.ts tests/structured-output.test.ts`
- ✅ `npm test` (`49` test files, `690` tests)

JackChen-me:
LGTM, merging.

JackChen-me (approved):

PR #181, opened 2026-04-26
Upstream/pr1 on plan ready

tizerluo:
## What
Adds an optional `onPlanReady` callback to `OrchestratorConfig` that fires after the coordinator decomposes a goal into tasks, before execution begins. Return `false` to abort the run.

## Why
Enables human-in-the-loop plan review before any agent work starts. Returning `false` aborts immediately with `{ success: false, agentResults: new Map() }`.

## Changes
- `src/types.ts` — add `onPlanReady?` to `OrchestratorConfig`
- `src/orchestrator/orchestrator.ts` — call hook before `executeQueue()`; update `Omit`/`Pick` types

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` passes (459 tests)
- [x] Only fires for `runTeam()` — `runAgent()` and `runTasks()` are unaffected

### DIFF
--- a/src/orchestrator/orchestrator.ts
+++ b/src/orchestrator/orchestrator.ts
@@ -814,8 +814,8 @@ async function buildTaskPrompt(task: Task, team: Team, queue: TaskQueue): Promis
  */
 export class OpenMultiAgent {
   private readonly config: Required<
-    Omit<OrchestratorConfig, 'onApproval' | 'onProgress' | 'onTrace' | 'defaultBaseURL' | 'defaultApiKey' | 'maxTokenBudget'>
-  > & Pick<OrchestratorConfig, 'onApproval' | 'onProgress' | 'onTrace' | 'defaultBaseURL' | 'defaultApiKey' | 'maxTokenBudget'>
+    Omit<OrchestratorConfig, 'onApproval' | 'onPlanReady' | 'onProgress' | 'onTrace' | 'defaultBaseURL' | 'defaultApiKey' | 'maxTokenBudget'>
+  > & Pick<OrchestratorConfig, 'onApproval' | 'onPlanReady' | 'onProgress' | 'onTrace' | 'defaultBaseURL' | 'defaultApiKey' | 'maxTokenBudget'>
 
   private readonly teams: Map<string, Team> = new Map()
   private completedTaskCount = 0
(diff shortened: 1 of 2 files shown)


JackChen-me:
Read through, this looks solid. Two things inline, both for parity with `onApproval`. Quick to address.

Heads up: branch is a few commits behind main and conflicts with the `maxTokenBudget` addition (#71 ). A rebase onto current main will sort it.

If you've got bandwidth, would be nice to have tests near `tests/approval.test.ts` for the reject and callback-throws paths. Not a blocker.

src/orchestrator/orchestrator.ts
```diff
@@ -714,6 +715,18 @@ export class OpenMultiAgent {
       runId,
     }
 
+    if (this.config.onPlanReady) {
```
JackChen-me, line 1179: Mirroring `onApproval`'s try/catch here would help (executeQueue ~728). As written, a thrown callback crashes the whole run, where `onApproval` catches it and skips remaining tasks as a controlled abort.

src/orchestrator/orchestrator.ts
```diff
@@ -714,6 +715,18 @@ export class OpenMultiAgent {
...
+      const tasks = queue.list()
+      const approved = await this.config.onPlanReady(tasks)
+      if (!approved) {
+        return {
```
JackChen-me, line 1183: Coordinator already burned tokens by the time this hook fires, so `totalTokenUsage: 0` under-reports actual cost. The other early-exit paths route through `buildTeamRunResult(...)` (e.g. ~1196) which preserves `cumulativeUsage`. Same pattern would work here.

JackChen-me:
Hey, rebased onto current main for you (just the `maxTokenBudget` `Omit/Pick` line, no behavior change). About both inline comments: going to fold the fixes into a small follow-up commit on main right after this lands, so no action needed from you. Squashing this in as-is. Thanks for the contribution!

src/orchestrator/orchestrator.ts
```diff
@@ -1175,6 +1176,18 @@ export class OpenMultiAgent {
...
+          success: false,
+          agentResults: new Map(),
+          totalTokenUsage: { input_tokens: 0, output_tokens: 0 },
```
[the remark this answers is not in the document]
JackChen-me: resolve #183

PR #182, opened 2026-04-26
Upstream/pr2 on agent stream

tizerluo:
## What
Adds an optional `onAgentStream` callback to `OrchestratorConfig` that fires for each `StreamEvent` during `runTeam()`. When set, `AgentPool` switches agents to streaming mode while preserving the concurrency semaphore.

## Why
Enables real-time streaming UIs without changing how `runTeam()` is called. Opt-in — callers without `onAgentStream` see no behavior change.

## Changes
- `src/types.ts` — add `onAgentStream?` to `OrchestratorConfig`
- `src/agent/pool.ts` — add `streamCallback` param; switch to `agent.stream()` when set
- `src/agent/agent.ts` — `stream()` accepts `Partial<RunOptions>`; merges timeout + caller abort signals
- `src/orchestrator/orchestrator.ts` — wire `onAgentStream` into `pool.run()`; update `Omit`/`Pick` types

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` passes (459 tests)
- [x] Only fires for `runTeam()` — `runAgent()` and `runTasks()` are unaffected

### DIFF
--- a/src/agent/agent.ts
+++ b/src/agent/agent.ts
@@ -239,13 +239,12 @@ export class Agent {
    *
    * Like {@link run}, this does not use or update the persistent history.
    */
-  // TODO(#18): accept optional RunOptions to forward trace context
-  async *stream(prompt: string): AsyncGenerator<StreamEvent> {
+  async *stream(prompt: string, runOptions?: Partial<RunOptions>): AsyncGenerator<StreamEvent> {
     const messages: LLMMessage[] = [
       { role: 'user', content: [{ type: 'text', text: prompt }] },
     ]
 
-    yield* this.executeStream(messages)
+    yield* this.executeStream(messages, runOptions)
   }
 
   // -------------------------------------------------------------------------
@@ -517,7 +516,7 @@ export class Agent {
    * Shared streaming path used by `stream`.
    * Handles state transitions and error wrapping.
    */
-  private async *executeStream(messages: LLMMessage[]): AsyncGenerator<StreamEvent> {
+  private async *executeStream(messages: LLMMessage[], callerOptions?: Partial<RunOptions>): AsyncGenerator<StreamEvent> {
     this.transitionTo('running')
 
     try {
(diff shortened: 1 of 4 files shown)


src/agent/agent.ts
```diff
@@ -496,8 +495,12 @@ export class Agent {
       const timeoutSignal = this.config.timeoutMs !== undefined && this.config.timeoutMs > 0
         ? AbortSignal.timeout(this.config.timeoutMs)
         : undefined
+      const callerAbort = callerOptions?.abortSignal
+      const effectiveAbort = timeoutSignal && callerAbort
+        ? mergeAbortSignals(timeoutSignal, callerAbort)
+        : timeoutSignal ?? callerAbort
 
-      for await (const event of runner.stream(messages, timeoutSignal ? { abortSignal: timeoutSignal } : {})) {
+      for await (const event of runner.stream(messages, effectiveAbort ? { abortSignal: effectiveAbort } : {})) {
```
JackChen-me, line 540: This branch only forwards `abortSignal`, but the orchestrator passes a few more things on `runOptions` (orchestrator.ts ~606): `onTrace`, `runId`, `traceAgent`, `team`. Dropping them silently breaks `onTrace` (no per-task or per-LLM-call traces fire when `onAgentStream` is set) and `delegate_to_agent` (any agent in streaming mode loses the team context, so the tool call fails).

One-line fix:

```ts
for await (const event of runner.stream(messages, { ...callerOptions, abortSignal: effectiveAbort })) { ... }
```

`runner.stream` already reads those fields from its options, so spreading them through is enough.

JackChen-me (commented):
Two things to address before merge, both flagged inline. The hook silently breaks `onTrace` and `delegate_to_agent` together (one-line fix). User callbacks that throw also trigger task retries via `executeWithRetry`, where `onTrace`'s `emitTrace` would have swallowed them.

Same rebase note as #181: branch is behind main, conflicts with `maxTokenBudget`. Rebase will fix it.

Same note on tests as #181, especially abort-during-stream since the signal merging is new.

src/agent/pool.ts
```diff
@@ -128,11 +128,27 @@ export class AgentPool {
...
+    streamCallback?: (event: StreamEvent) => void,
   ): Promise<AgentRunResult> {
     const agent = this.requireAgent(agentName)
 
     await this.semaphore.acquire()
     try {
+      if (streamCallback) {
+        let result: AgentRunResult | null = null
+        for await (const event of agent.stream(prompt, runOptions)) {
+          streamCallback(event)
```
JackChen-me: `streamCallback(event)` is called raw here. If a user's callback throws, the error escapes `pool.run` and gets caught by `executeWithRetry` (orchestrator.ts:265), which retries the entire task. Each retry hits the same callback bug and burns another LLM call until `maxRetries` is exhausted.

`emitTrace` (utils/trace.ts) enforces the opposite contract for `onTrace`: callback errors are intentionally swallowed because observability must not break execution. `onAgentStream` is the same shape of subscription and probably wants the same guarantee.

```ts
try {
  streamCallback(event)
} catch {
  // streaming callback must not affect execution
}
```

Worth mirroring `emitTrace`'s swallow pattern, unless the throw-propagates behavior is intentional?

PR #183, opened 2026-04-26
fix(orchestrator): polish onPlanReady abort path

JackChen-me:
## Summary

Follow-up to #181. Fixes three issues in the `onPlanReady` abort path:

1. **Token reporting:** the abort early-return was using hard-coded `totalTokenUsage: 0`, which under-reports actual cost since the coordinator decomposition has already burned tokens. Now routes through `buildTeamRunResult` and overrides `success: false`, so callers tracking spend/quotas see real usage while still getting the abort signal. (Codex flagged this as P1 on #181.)
2. **Thrown callbacks:** wraps the `onPlanReady` invocation in `try/catch` to match `onApproval`'s pattern. A thrown callback is now a controlled abort instead of crashing the entire run.
3. **Parameter type:** tightens `(tasks: Task[])` to `(tasks: readonly Task[])` for consistency with `onApproval`. This is technically source-breaking under `strictFunctionTypes` for any caller that explicitly typed their callback as `(tasks: Task[]) => ...`. Calling out the timing: `onPlanReady` was added in #181 and is not yet on npm (latest published is 1.3.0; #181 will ship in 1.4.0). Tightening before the initial release avoids a future semver-major to align with `onApproval`'s `readonly` parameters.

JSDoc on `onPlanReady` is expanded to document the abort semantics and the no-mutation rule.

## Test plan

- [x] `npm run lint` passes
- [x] `npm test` passes (694 tests, +3 new)
- [x] New tests cover the three abort paths (`returns false`, `throws`, `returns true → proceeds`) and assert that the result preserves coordinator decompose token usage on abort

### DIFF
--- a/src/orchestrator/orchestrator.ts
+++ b/src/orchestrator/orchestrator.ts
@@ -1177,14 +1177,15 @@ export class OpenMultiAgent {
     }
 
     if (this.config.onPlanReady) {
-      const tasks = queue.list()
-      const approved = await this.config.onPlanReady(tasks)
+      const planTasks = queue.list()
+      let approved: boolean
+      try {
+        approved = await this.config.onPlanReady(planTasks)
+      } catch {
+        return { ...this.buildTeamRunResult(agentResults, goal, []), success: false }
+      }
       if (!approved) {
-        return {
-          success: false,
-          agentResults: new Map(),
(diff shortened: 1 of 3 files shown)


JackChen-me:
@codex

PR #188, opened 2026-04-27
feat: add plan_ready and agent_stream as TraceEvent types

ibrahimkzmv:
## What

Introduces TraceEvent types (`plan_ready`, `agent_stream`) for the observation layer.

## Why

To enable observability and streaming visibility into the system

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -187,6 +187,8 @@ export type {
   ToolCallTrace,
   TaskTrace,
   AgentTrace,
+  PlanReadyTrace,
+  AgentStreamTrace,
 
   // Memory
   MemoryEntry,
--- a/src/orchestrator/orchestrator.ts
+++ b/src/orchestrator/orchestrator.ts
@@ -1199,44 +1199,31 @@ export class OpenMultiAgent {
       taskMetrics,
     }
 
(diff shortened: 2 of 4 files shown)


JackChen-me (commented):
Two things compound here. Extending the exported `TraceEventType` / `TraceEvent` unions at `src/types.ts:735` is semver-significant for TS consumers with exhaustive handlers (`switch` with a return value, `assertNever`, `Record<TraceEventType, ...>`, etc.). Separately, `plan_ready` at `src/orchestrator/orchestrator.ts:1202` only fires when `onPlanReady` is installed, so ordinary `onTrace` callers never see the event the new type advertises. As written, we pay the compatibility cost without delivering the full observability benefit.

Suggested fix: emit `plan_ready` unconditionally at the plan boundary after `queue.list()`. Default `approved` to `true`, override it with the `onPlanReady` result when the callback is present, and treat callback throws as `approved: false`. If we do that, this becomes much easier to justify shipping with a release note.

ibrahimkzmv:
You are right. I've just made the changes.

PR #191, opened 2026-04-29
Example: customer ticket

### LINKED ISSUE #133: Example: customer support ticket pipeline (Express route integration)
## Context

Discussion #33 survey shows "customer support / chatbot systems" is one of the most-requested use cases. OMA can run inside an existing Node backend, which Python-based frameworks can't. Need an example that makes the embedded-in-backend story concrete.

## Task

Add under `examples/integrations/express-customer-support/`:

- `index.ts`: Express app with `POST /tickets` route
- `README.md`: curl command + expected output
- `package.json`: pins `express` as a devDependency so the root project doesn't take it on

The route receives `{ subject, body }`. `runTasks()` executes an explicit three-stage DAG: classifier (urgency + category) → drafter (reply, depends on classifier) → QA reviewer (tone + accuracy, depends on drafter). Response returns structured `{ category, urgency, draft_reply, qa_notes }`.

## Reference

- `examples/patterns/structured-output.ts`: Zod output schema
- `examples/basics/task-pipeline.ts`: runTasks DAG pattern

## Acceptance

- Express server runs on a local port (default 3000, overridable via env)
- One curl command in the example README triggers a full pipeline run and returns valid JSON matching the Zod schema
- Agents use different models (classifier cheap, drafter mid, QA strong), parameterized via env so free-tier providers work
- Error paths handled: invalid JSON body returns 400; LLM failure returns 502 with structured error; timeout > 60s returns 504
- Smoke test included (`supertest` or plain `fetch`) that runs the pipeline once and asserts response shape

## Notes

Keep the Express setup minimal. No middleware bloat. The point is showing OMA inside a real backend, not writing an Express tutorial.

CodingBangboo:
## What

Fixes #133 
Adds a self-contained example under `examples/integrations/express-customer-support/` showing OMA embedded in an Express backend. Per the discussion on the issue, the pipeline uses `runTasks()` (fixed three-stage DAG) rather than `runTeam()`.

  ## What each file does

  - **`index.ts`** — Express app exposing `POST /tickets`. Defines three agents (classifier → drafter → qa-reviewer), each with its own Zod `outputSchema`, and runs them as a `runTasks()` DAG on every request. Includes a `pickAgent()` helper that resolves provider+model from env vars (`CLASSIFIER_PROVIDER`/`_MODEL`, etc.) and fails fast at startup if the matching API key is missing. HTTP error mapping: 400 on invalid JSON body, 502 on
  pipeline failure or missing structured output, 504 on >60s timeout (wired through an `AbortController` so in-flight LLM calls are cancelled). Exports a
  `createApp()` factory and the response Zod schema; only auto-starts the listener when run directly.

  - **`smoke.ts`** — Imports `createApp()`, binds an ephemeral port, sends one real `POST /tickets`, and validates the response against the exported
  `SupportTicketResponse` schema. Exits 0 on success, 1 on failure. Requires `ANTHROPIC_API_KEY` (or whatever provider the user has configured).

  - **`README.md`** — Setup, single curl command + expected response shape, env var reference (per-agent provider/model overrides), HTTP error matrix, and "Swapping providers" section showing how to mix providers per tier (e.g., Gemini free-tier classifier + Anthropic drafter/QA).

  - **`package.json`** — Pins `express` and `tsx` as devDependencies so the root project doesn't take them on. `prestart` rebuilds the parent repo before `npm start`, matching the convention in `with-vercel-ai-sdk`.

  - **`package-lock.json`** — Lockfile for the example's deps.

  - **`examples/README.md`** — One-row addition under the `integrations` table pointing to the new example.
## Why

Fixes #133

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -69,6 +69,7 @@ Hooking the framework up to outside-the-box tooling.
 | [`integrations/trace-observability`](integrations/trace-observability.ts) | `onTrace` spans for LLM calls, tools, and tasks. |
 | [`integrations/mcp-github`](integrations/mcp-github.ts) | An MCP server's tools exposed to an agent via `connectMCPTools()`. |
 | [`integrations/with-vercel-ai-sdk/`](integrations/with-vercel-ai-sdk/) | Next.js app — OMA `runTeam()` + AI SDK `useChat` streaming. |
+| [`integrations/express-customer-support/`](integrations/express-customer-support/) | Express REST API — `runTasks()` behind POST /tickets with per-agent Zod output schemas, mix-and-match provider env vars, and HTTP error mapping (400/502/504). |
 
 ## production — real-world use cases
 
--- a/examples/integrations/express-customer-support/README.md
+++ b/examples/integrations/express-customer-support/README.md
@@ -1,0 +1,103 @@
+# Express Customer Support
+
+An Express REST API that wraps OMA's `runTeam()` coordinator pattern behind a single `POST /tickets` endpoint. A three-agent pipeline runs on every request:
+
+1. **classifier** (`claude-haiku-4-5-20251001`) — categorises the ticket and assigns urgency
+2. **drafter** (`claude-sonnet-4-6`) — writes a polished customer-facing reply
+3. **qa-reviewer** (`claude-opus-4-7`) — reviews the draft for tone and accuracy
+
+Each agent uses a Zod `outputSchema`; the endpoint assembles the three structured results into one JSON response. Each agent's model **and** provider are independently swappable via env vars so a free-tier setup can mix providers (see [Swapping providers](#swapping-providers) below).
+
+## Setup
+
+```bash
+cd examples/integrations/express-customer-support
+npm install
+export ANTHROPIC_API_KEY=sk-ant-...    # default — all three agents use Anthropic
+```
+
+## Start the server
+
+```bash
+npm start
+# Support API listening on http://localhost:3000
+```
+
+## Send a ticket
+
+```bash
+curl -s -X POST http://localhost:3000/tickets \
+  -H 'Content-Type: application/json' \
+  -d '{"subject":"My order never arrived","body":"I placed order #12345 two weeks ago and it still has not shipped. Please help!"}' | jq .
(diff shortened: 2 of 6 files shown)


JackChen-me:
Two README things to fix before merge:

**L9.** "via the shared-memory dependency-context block" mixes two mechanisms. Context comes from `queue.complete(task.id, result.output)` and `buildTaskPrompt` (the `## Context from prerequisite tasks` block), not from shared memory. Shared memory only feeds a prompt under `memoryScope: 'all'`. Drop the "shared-memory" framing.

**L80.** Supported `*_PROVIDER` list misses `copilot`, `minimax`, `azure-openai`, all of which are in `PROVIDER_ENV_KEYS`. Also "Other providers ... will also work" hides a footgun: unmapped providers silently skip the startup check (`if (envKey && ...)`), so a missing key shows up as a 502 mid-request rather than at startup. Either expand the map or call out the gap.

Also `sharedMemory: true` is incidental here. `runTasks` builds dependency context from `task.result`, not from memory. Either drop it or note it's audit-only.

CodingBangboo:
Thanks for reviewing! @JackChen-me
Let me know how it looks now.

> Two README things to fix before merge:
> 
> **L9.** "via the shared-memory dependency-context block" mixes two mechanisms. Context comes from `queue.complete(task.id, result.output)` and `buildTaskPrompt` (the `## Context from prerequisite tasks` block), not from shared memory. Shared memory only feeds a prompt under `memoryScope: 'all'`. Drop the "shared-memory" framing.
> 
Fixed.
> **L80.** Supported `*_PROVIDER` list misses `copilot`, `minimax`, `azure-openai`, all of which are in `PROVIDER_ENV_KEYS`. Also "Other providers ... will also work" hides a footgun: unmapped providers silently skip the startup check (`if (envKey && ...)`), so a missing key shows up as a 502 mid-request rather than at startup. Either expand the map or call out the gap.
> 
Added the providers and also called out the gap.
> Also `sharedMemory: true` is incidental here. `runTasks` builds dependency context from `task.result`, not from memory. Either drop it or note it's audit-only.
> 
Removed.

PR #193, opened 2026-04-30
fix(runner): preserve tool_use/tool_result pairs in sliding-window truncation

MyPrototypeWhat:
## Problem

`AgentRunner.truncateToSlidingWindow` slices raw message count (`afterFirst.slice(-maxTurns * 2)`), assuming "2 messages = 1 user/assistant pair". Tool round-trips violate that assumption: an `assistant` message carrying `tool_use` blocks is followed by a `user` message carrying `tool_result` blocks — the role order is reversed compared to a plain conversation pair.

When the slice boundary lands on a `tool_result`, the matching `tool_use` upstream gets dropped. The next chat call then ships an orphan `tool_use_id` that Anthropic and OpenAI APIs reject.

## Repro (added as a regression test)

```
history = [user, asst_tool_use, user_tool_result, asst_text, user_text, asst_text]
contextStrategy: { type: 'sliding-window', maxTurns: 2 }
```

Pre-fix: `afterFirst.slice(-4) = [user_tool_result, asst_text, user_text, asst_text]` → `tool_result(tu-1)` reaches the adapter without its matching `tool_use`.

Verified locally with a standalone mock-adapter demo: pre-fix output shows `🚨 ORPHAN tool_result_ids: ['tu-1']`; post-fix shows `✅ Every tool_result has a matching tool_use`.

## Fix

New top-level `groupIntoTurns(messages)` helper fuses each `assistant`-with-`tool_use` together with its following `user`-with-`tool_result` into one atomic turn. `truncateToSlidingWindow` now slices on turn boundaries instead of message indices, so tool pairs stay intact (or get dropped together).

The same primitive solves the same class of bug in [context-chef](https://github.com/MyPrototypeWhat/context-chef)'s Janitor module (`packages/core/src/modules/janitor/index.ts:28-51`) — referenced as prior art in the source comment.

## Test plan

- [x] New regression test (`sliding-window keeps tool_use/tool_result pairs together`) fails on `main`, passes after the fix
- [x] Existing 15 context-strategy tests still pass
- [x] Full suite: 707/707 passed
- [x] `tsc --noEmit` clean
- [x] No `package-lock.json` noise

### DIFF
--- a/src/agent/runner.ts
+++ b/src/agent/runner.ts
@@ -210,6 +210,56 @@ function extractToolUseBlocks(content: readonly ContentBlock[]): ToolUseBlock[] 
   return content.filter((b): b is ToolUseBlock => b.type === 'tool_use')
 }
 
+/**
+ * Boundaries (`startIndex` inclusive, `endIndex` exclusive) of a single
+ * atomic conversation turn within a flat message array.
+ */
+interface Turn {
+  startIndex: number
+  endIndex: number
+}
+
+/**
+ * Group a flat message array into atomic turns so context-management
+ * strategies can split on safe boundaries.
+ *
+ * A turn is one of:
+ *   - a single user / assistant text message, or
+ *   - an assistant message containing one or more `tool_use` blocks plus the
+ *     immediately following user message containing the matching `tool_result`
+ *     blocks (kept together so neither half can be dropped on its own).
+ *
+ * Splitting on turn boundaries — instead of slicing by raw message count —
+ * prevents orphaned `tool_use_id` references that the Anthropic and OpenAI
+ * APIs reject. Modelled on `groupIntoTurns` from the context-chef library.
+ */
+function groupIntoTurns(messages: LLMMessage[]): Turn[] {
+  const turns: Turn[] = []
+  let i = 0
+  while (i < messages.length) {
+    const msg = messages[i]!
+    const hasToolUse =
+      msg.role === 'assistant' && msg.content.some(b => b.type === 'tool_use')
+    if (hasToolUse) {
+      const start = i
+      i++
+      // Absorb the matching tool_result user message, when present.
+      if (
+        i < messages.length
+        && messages[i]!.role === 'user'
+        && messages[i]!.content.some(b => b.type === 'tool_result')
+      ) {
+        i++
+      }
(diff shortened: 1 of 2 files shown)


MyPrototypeWhat:
### Semantic change to disclose

Surfacing this proactively before review so we can align on the intended behavior.

This fix shifts the unit of `maxTurns` for `sliding-window` from **"message pair"** (the pre-fix de-facto behavior, since the implementation sliced `messages * 2`) to **"atomic turn"** (post-fix, since `groupIntoTurns` is the new boundary).

| History shape                                    | pre-fix `maxTurns=1`            | post-fix `maxTurns=1`                  |
| ------------------------------------------------ | ------------------------------- | -------------------------------------- |
| Plain text `[u, a, u, a, u, a]`                  | keeps last **2 msgs** (1 pair)  | keeps last **1 msg** (1 turn)          |
| Tool round-trip `[u, a_tool, u_res, a_tool, u_res]` | keeps last 2 (could split pair) | keeps last 1 atomic turn = **2 msgs** intact |

### Implications

- For plain-text histories this is technically a **behavior change**: `maxTurns=10` previously yielded ~20 messages, now yields ~10.
- For tool-heavy histories the previous count was always approximate due to the bug — now it's exact and pair-safe.
- The current `ContextStrategy` type (`{ type: 'sliding-window'; maxTurns: number }` in `src/types.ts`) does not document the unit, so the public contract is not broken — but I want to make sure this matches your intent before merging.

### Options

1. **Land as-is** + add a one-line jsdoc to `maxTurns` in `types.ts` clarifying it counts atomic turns. (My preference — the new semantics matches the field name more honestly.)
2. **Preserve message-pair count semantics** — adjust `truncateToSlidingWindow` to walk turns from the tail accumulating until message count ≥ `maxTurns * 2`. Same bug fix, but `maxTurns` keeps its old effective unit.

Happy to follow up with whichever you prefer. Option 1 is a 1-line doc addition; option 2 is ~10 lines of additional logic.

JackChen-me:
I’d prefer option 2 here. Please preserve the existing effective `sliding-window.maxTurns` behavior for plain conversations, and fix only the `tool_use/tool_result` truncation bug in this PR.

PR #194, opened 2026-04-30
Add aws bedrock adaptor

### LINKED ISSUE #23: [Feature] AWS Bedrock LLMAdapter
Add support for AWS Bedrock as a meta-provider, enabling access to Claude, Llama, Mistral, and other models through a unified AWS endpoint.

Useful for enterprise users who route all LLM traffic through their AWS account.

**Implementation:**
- Add `src/llm/bedrock.ts` using `@aws-sdk/client-bedrock-runtime`
- Register `'bedrock'` in `createAdapter()` factory
- Add example to `examples/`

Auth: AWS credentials (access key / IAM role)

CodingBangboo:
## What

Fixes #23 
- Adds `BedrockAdapter` (`src/llm/bedrock.ts`) backed by the Bedrock Converse / ConverseStream APIs, supporting all Bedrock-hosted model families (Claude, Llama, Mistral, Cohere, Titan) through a single unified interface
- Registers `'bedrock'` in the `createAdapter()` factory with lazy import and a `region` parameter; credentials resolve via the AWS SDK default provider chain (env vars, shared config, IAM role)
- Adds `@aws-sdk/client-bedrock-runtime` as an optional peer dependency — users only install it if they need Bedrock, preserving the three-runtime-dependency promise for `dependencies`

## What's included

- `src/llm/bedrock.ts` — LLMAdapter implementation: `chat()`, `stream()`, message/tool/image conversion, abortSignal support, reasoning output captured; **limit**: multi-turn tool-use round-trip with extended thinking is not yet supported (Bedrock requires the reasoning text + signature to be echoed back, which the framework's yet to have the shape for `ReasoningBlock.signature` )
- `src/llm/adapter.ts` — `'bedrock'` added to `SupportedProvider` union and factory
- `src/cli/oma.ts` — provider reference entry and default model hint (`anthropic.claude-3-5-haiku-20241022-v1:0`, an on-demand-throughput ID that works without cross-region inference profile prefix)
- `examples/providers/bedrock.ts` — three-agent `runTasks()` pipeline demo with setup/auth notes
- `tests/bedrock-adapter.test.ts` — 21 unit tests covering chat, stream, tool use, reasoning, error paths, and region resolution
- Docs: `README.md`, `README_zh.md`, `examples/README.md`, `CLAUDE.md` updated

## Why

Fixes #23 

## Note on model IDs

Newer Claude models on Bedrock (Sonnet 4.x, Haiku 4.x) require a cross-region inference profile prefix (e.g. `us.anthropic.claude-haiku-4-5-20251001-v1:0`) rather than a bare model ID. The README and CLI hint use `anthropic.claude-3-5-haiku-20241022-v1:0` as the suggested default since it supports on-demand throughput without the prefix. The example file comments explain the distinction.

## Note on `package-lock.json`

The lockfile grew by ~2400 lines. This is expected — `@aws-sdk/client-bedrock-runtime` is added as a `devDependency` for testing and brings in several `@aws-sdk/*` sub-packages. End users who don't need Bedrock are unaffected since it remains an optional peer dependency and is not installed transitively.


JackChen-me:
Left 2 comments. One is a direct bug: `stream()` yields reasoning deltas but drops them from the final `done` payload, so the streaming contract doesn't match `chat()` or the other adapters.

The reasoning round-trip issue is broader than this PR, but it's still real here. Bedrock requires reasoning text + signature to be sent back unchanged in tool-use / multi-turn flows, and the current framework shape can't preserve that, so this adapter can't fully support that path yet. Fine to split the larger type/API cleanup into follow-up work, but then this PR should narrow the claim instead of implying full reasoning support.

CodingBangboo:
@JackChen-me Hi Jack, your comments in this conversation look good to me and I will be working on them. Just wanted to double check

> Left 2 comments.

Am I missing any inline review comments or the above comment is everything?

Thanks

JackChen-me:
Yeah, the above comment is everything. @CodingBangboo

CodingBangboo:
> One is a direct bug: `stream()` yields reasoning deltas but drops them from the final `done` payload, so the streaming contract doesn't match `chat()` or the other adapters.

Has fixed this bug and also kept honest limitation comments in commits.


> The reasoning round-trip issue is broader than this PR, but it's still real here. Bedrock requires reasoning text + signature to be sent back unchanged in tool-use / multi-turn flows, and the current framework shape can't preserve that, so this adapter can't fully support that path yet. Fine to split the larger type/API cleanup into follow-up work, but then this PR should narrow the claim instead of implying full reasoning support.

Has relaxed the description on reasoning block support for the current PR and pointed out the limitation.

**Other notes**:
Azure Open AI adaptor and co-pilot seem to have similar bugs in this PR. Worth a manual check.

Shall we resolve all similar bugs in a separate PR or we can have an issue to track these?

Open to any other comments before merging.
@JackChen-me

PR #197, opened 2026-05-01
feat(trace): include tool input/output in ToolCallTrace (closes #124)

### LINKED ISSUE #124: [Feature] Fix to allow trace to contain input/output details of tool calls.
# Problem

TraceEvent allows us to see when a task starts or a tool is called, but if you dig any deeper you will find that the tool call doesn't allow you to see what was sent and returned. I can see all this information in LM Studio server logs but this is inconvenient.

# Solution

In `runner.ts` we just make the `emitTrace` the last thing that happens, and pass it the new input/output

          const record: ToolCallRecord = {
            toolName: block.name,
            input: block.input,
            output: result.data,
            duration,
          }
          const resultBlock: ToolResultBlock = {
            type: 'tool_result',
            tool_use_id: block.id,
            content: result.data,
            is_error: result.isError,
          }
          if (options.onTrace) {
            emitTrace(options.onTrace, {
              type: 'tool_call',
              runId: options.runId ?? '',
              taskId: options.taskId,
              agent: options.traceAgent ?? this.options.agentName ?? 'unknown',
              tool: block.name,
              isError: result.isError ?? false,
              input: block.input,
              output: result.data,
              startMs: startTime,
              endMs: endTime,
              durationMs: duration,
            })
          }

We also have to make a change in `export interface ToolCallTrace extends TraceEventBase`
`
readonly input: Record<string, unknown>
  readonly output: string
`

Now we can use it to capture tools
```

export function handleTrace(event: TraceEvent): void {
    case 'tool_call': {
      const toolStatus = event.isError ? '\n[ERROR]';
      console.log(`[TOOL EXECUTED] ran '${event.tool}');

      const inputStr = JSON.stringify(event.input);
      const outStr = String(event.output);
      
      toolExecutionHistory.add(event.tool);

      switch (event.tool) {
        case 'custom_tool_name': { break; }

```

MyPrototypeWhat:
## What this does

`ToolCallTrace` previously surfaced only `tool`, `isError`, and `durationMs`. To diagnose a tool call you had to dig through provider server logs (LM Studio, vLLM) or wire up `onToolCall` / `onToolResult` as a parallel channel — exactly what `onTrace` is supposed to replace.

This PR adds two readonly fields to `ToolCallTrace`:

- `input: Record<string, unknown>` — mirrors the originating `ToolUseBlock.input`
- `output: string` — mirrors `ToolResult.data`, including any executor truncation (per-tool `maxOutputChars` / agent `maxToolOutputChars`) so trace consumers see exactly what the LLM will see in the next turn

A single `onTrace` callback now sees the full tool-call lifecycle.

Closes #124.

## Scope

| File | Change |
|------|-------|
| `src/types.ts` | `ToolCallTrace` gains `input` + `output` (both `readonly`) with jsdoc |
| `src/agent/runner.ts` | `emitTrace('tool_call', ...)` populates the two new fields with `block.input` and `result.data` |
| `tests/trace.test.ts` | Extended success + error tests assert the new fields; new regression `tool_call trace.output reflects executor truncation` pins the jsdoc contract |
| `examples/integrations/trace-observability.ts` | `[TOOL]` log line shows truncated input/output for readability; comment notes the full payload remains on the event |

## Behaviour preserved

- **Pure type-additive**: code that *reads* trace events keeps working — the new fields are opt-in by access
- **Single source of truth for output length**: `ToolExecutor.execute` is where truncation happens; `output` reflects the same string the LLM sees on the next turn
- **No new public API**: no orchestrator option, no callback shape change

## Type-level note

These are required `readonly` fields, so code that *constructs* a `ToolCallTrace` literally — e.g. test mocks, trace proxies — must populate `input` and `output`. The fix is mechanical and tsc points straight at the call site. Code that only reads trace events (the 99% case) is unaffected.

Other `TraceEvent` variants are unchanged.

## Test plan

- [x] Extended `emits tool_call trace with correct fields` and `tool_call trace has isError: true on tool failure` to assert the new \`input\` / \`output\` fields
- [x] New regression \`tool_call trace.output reflects executor truncation\` pins the contract that per-tool \`maxOutputChars\` is reflected in the trace
- [x] Full suite: 710/710 pass
- [x] \`tsc --noEmit\` clean

### DIFF
--- a/examples/integrations/trace-observability.ts
+++ b/examples/integrations/trace-observability.ts
@@ -42,6 +42,11 @@ const writer: AgentConfig = {
 // Trace handler — log every span with timing
 // ---------------------------------------------------------------------------
 
+/** Truncate long strings for log readability — full payload stays in the event. */
+function truncate(s: string, max = 80): string {
+  return s.length > max ? `${s.slice(0, max)}…(+${s.length - max} chars)` : s
+}
+
 function handleTrace(event: TraceEvent): void {
   const dur = `${event.durationMs}ms`.padStart(7)
 
@@ -53,9 +58,14 @@ function handleTrace(event: TraceEvent): void {
       )
       break
     case 'tool_call':
+      // Truncation here is for log readability only — the full payload remains
+      // on `event.input` / `event.output` for downstream consumers (OpenTelemetry,
+      // structured log shippers, audit pipelines, etc.).
       console.log(
         `  [TOOL]  ${dur}  agent=${event.agent}  tool=${event.tool}` +
-        `  error=${event.isError}`,
+        `  error=${event.isError}` +
+        `  input=${truncate(JSON.stringify(event.input))}` +
+        `  output=${truncate(event.output)}`,
       )
       break
     case 'task':
--- a/src/agent/runner.ts
+++ b/src/agent/runner.ts
@@ -920,6 +920,8 @@ export class AgentRunner {
               agent: options.traceAgent ?? this.options.agentName ?? 'unknown',
               tool: block.name,
               isError: result.isError ?? false,
+              input: block.input,
+              output: result.data,
               startMs: startTime,
               endMs: endTime,
               durationMs: duration,
--- a/src/types.ts
+++ b/src/types.ts
@@ -779,6 +779,10 @@ export interface ToolCallTrace extends TraceEventBase {
   readonly type: 'tool_call'
   readonly tool: string
   readonly isError: boolean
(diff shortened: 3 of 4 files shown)


JackChen-me (approved):
Clean. One asymmetry worth flagging: `output` is executor-truncated but `input` isn't (e.g. `file_write` content goes in raw). Not blocking since `onTrace` is opt-in, can add `maxTraceInputChars` if it comes up.

PR #203, opened 2026-05-02
[Feature]PlanOnly mode

### LINKED ISSUE #192: [Feature] planOnly mode for runTeam() — inspect task decomposition without executing
## Source

**Where did this idea come from?** (Pick one — helps maintainers triage and prioritize.)

- [ ] **Real use case** — I'm using open-multi-agent and hit this limit. Describe the use case in "Problem" below.

## Problem

When building against `runTeam()`, there are two recurring needs that the current API makes awkward:

**1. Integration testing.** I want to assert the coordinator's task decomposition (task count, titles, assignees, dependency graph) without spending execution tokens. Today there's no clean way to do this — the only option is to run the whole team and check results after the fact.

**2. Debugging and cost estimation.** Before committing a long multi-agent run, it's useful to print "here are the tasks the coordinator will create" as a dry-run check.

The existing `onPlanReady` callback (#181 ) can approximate this — return `false` to abort before execution starts. But it has two problems in this use case (it is correct on its own):

- It returns `success: false`, which is **semantically** wrong. Nothing failed; we just wanted to inspect the plan. Any code that checks `result.success` will treat an intentional plan-only call as an error.
- It's imperative (wire a callback) rather than declarative (pass a flag). A test or debug helper has to construct a callback just to get plan-only behavior.

## Proposed Solution

Add a `planOnly` option to the third argument of `runTeam()`

The coordinator runs normally and decomposes the goal. Execution stops before any task agent starts. The returned `TeamRunResult` has `success: true` and `tasks` populated, which is already part of the existing type — no type changes needed beyond `TeamRunResult` documenting this as the valid plan-only shape.

Internally, this is a short-circuit in `orchestrator.ts` right after the coordinator finishes and before `executeQueue()` is called — the same boundary where `onPlanReady` fires today.

## Alternatives Considered

**`onPlanReady` returning `false`** — works mechanically but is semantically broken for this use case (`success: false` implies a crash or abort, not a deliberate inspection). It also requires the caller to store the task list inside the callback closure; the return value of `runTeam()` has no clean way to surface the tasks in the abort path.

**Post-run inspection** — run the full team and read `result.tasks`. Costs execution tokens. Defeats the purpose for testing and estimation.

## Additional context

Add any other context, code examples, or screenshots about the feature request here.
The implementation surface is narrow: one new optional boolean on the `runTeam` options type, and a guard in `orchestrator.ts` after the coordinator completes. The `onPlanReady` path already marks this boundary cleanly — `planOnly` would just codify "stop here and return success" as a first-class mode instead of a callback workaround.

## Questions:
- Should `planOnly` live on `OrchestratorConfig` (affects all calls on this instance) or on the `runTeam()` third argument (per-call)?
- Does `TeamRunResult` need a new field (e.g. `planOnly: true`) to let callers distinguish a plan-only result from a full run? Or is `agentResults.size === 0` sufficient?

Happy to implement this if the direction looks right.

CodingBangboo:
## What

Adds a `planOnly` option to `runTeam()` that lets the coordinator decompose a goal into tasks without executing any task agents. The result has `success: true`, `planOnly: true`, and `tasks` populated with pending/blocked task records: useful for integration testing the task DAG or previewing cost before a real run.

## What changed

- **`src/types.ts`** — new `RunTeamOptions` interface (replaces the inline anonymous options type on `runTeam()`); adds `planOnly?: boolean` to both `RunTeamOptions` and `TeamRunResult`
- **`src/orchestrator/orchestrator.ts`** — `runTeam()` now accepts `RunTeamOptions`; bypasses the simple-goal short circuit when `planOnly: true`; inserts a short-circuit after `onPlanReady` fires (but before `executeQueue`) that returns `success: true` with tasks from `queue.list()`
- **`src/index.ts`** — exports `RunTeamOptions` as part of the public API
- **`tests/orchestrator.test.ts`** — four new test cases covering: plan returned without executing tasks, `onPlanReady` still fires, `onPlanReady` rejection wins over `planOnly`, and simple-goal short-circuit is bypassed

## Behavior notes

- `onPlanReady` still fires when `planOnly: true`. If the callback returns `false`, the rejection wins — result is `success: false` and `planOnly` is `undefined` (same as today's abort behavior).
- `planOnly` is a per-call option on the third argument, not on `OrchestratorConfig`: consistent with the scope guidance in the issue.
- The simple-goal short-circuit is bypassed so the coordinator always runs, which means `planOnly` on a single-agent team is valid and returns a single-task plan.

## Why

fixed #192

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,18 @@ console.log(`Tokens: ${result.totalTokenUsage.output_tokens} output tokens`)
 
 For MapReduce-style fan-out without task dependencies, use `AgentPool.runParallel()` directly. See [`patterns/fan-out-aggregate`](examples/patterns/fan-out-aggregate.ts).
 
+### Preview the plan without running it
+
+Pass `planOnly: true` to `runTeam()` and the coordinator decomposes the goal but no task agents execute. The result carries `planOnly: true`, populated `tasks` (no metrics), and only the coordinator's token usage. Useful for asserting on the task DAG in integration tests, or previewing cost before a real run.
+
+```ts
+const plan = await orchestrator.runTeam(team, goal, { planOnly: true })
+
+for (const task of plan.tasks ?? []) {
+  console.log(`${task.title} → ${task.assignee} (deps: ${task.dependsOn.length})`)
+}
+```
+
 ### Run from the shell
 
 For shell and CI, the package exposes a JSON-first binary. See [docs/cli.md](./docs/cli.md) for `oma run`, `oma task`, `oma provider`, exit codes, and file formats.
--- a/README_zh.md
+++ b/README_zh.md
@@ -148,6 +148,18 @@ console.log(`Tokens: ${result.totalTokenUsage.output_tokens} output tokens`)
 
 要 MapReduce 风格的 fan-out 但不需要任务依赖，直接用 `AgentPool.runParallel()`。例子见 [`patterns/fan-out-aggregate`](examples/patterns/fan-out-aggregate.ts)。
 
+### 只预览计划，不真跑
+
+给 `runTeam()` 传 `{ planOnly: true }`，coordinator会把目标拆成任务，但不执行任何 agent。结果里会带上 `planOnly: true`、完整的 `tasks`（不带 metrics），以及仅coordinator部分的 token 消耗。适合在integration tests里assert任务 DAG，或在真跑之前先看看成本。
+
+```ts
+const plan = await orchestrator.runTeam(team, goal, { planOnly: true })
+
+for (const task of plan.tasks ?? []) {
+  console.log(`${task.title} → ${task.assignee}（依赖：${task.dependsOn.length}）`)
+}
+```
+
 ### 从命令行运行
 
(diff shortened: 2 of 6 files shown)


JackChen-me:
@codex review plz

JackChen-me:
Approach looks right, but there is one real lifecycle bug.

P1 (must fix): `agent_start` fires for the coordinator before decomposition, but the new `planOnly` path returns before the matching `agent_complete`, which currently only fires after synthesis. Anyone tracking active agents via `onProgress` will see the coordinator as permanently running for successful plan-only runs. Small fix: emit `agent_complete` for the decomposition result before the `planOnly` return, and add a regression test that `agent_start` / `agent_complete` stay balanced.

P2 (optional, can be follow-up): `renderLiveOutput` in the dashboard treats only `completed`, `failed`, `skipped`, and `blocked` as terminal. In a plan-only result, root tasks intentionally remain `pending`, so the dashboard header says "Task graph execution in progress" instead of showing this as a completed plan preview. This is outside the main `runTeam` path, so I would not block on it, but it is worth fixing for callers who render plan-only results.

JackChen-me:
P1 fixed. P2 is non-blocking / follow-up. Good to merge.

Issue #200, opened 2026-05-02
[Proposal] Unified reasoning preservation across Anthropic / OpenAI / Gemini adapters


MyPrototypeWhat:
## Motivation

OMA's IR (`LLMMessage.content` → `ContentBlock`) defines `ReasoningBlock` (`src/types.ts:22`), but inspecting all three native adapters reveals that **none of them correctly preserve reasoning artifacts across multi-turn calls**:

| Adapter | Incoming text | Incoming signature | Outgoing | Request param |
|---------|--------------|---------------------|----------|---------------|
| Anthropic | ✅ `anthropic.ts:164-169` | ❌ signature dropped | ❌ stripped | ❌ no `thinking` |
| OpenAI | ✅ `openai-common.ts:228-231` | N/A (no protocol) | ❌ silent skip (API enforces) | ❌ no `reasoning_effort` |
| Gemini | ❌ thought parts ignored | ❌ thoughtSignature dropped | ❌ stripped | ❌ no `thinkingConfig` |

This breaks each provider's multi-turn reasoning protocol differently:

- **Anthropic**: per [extended thinking docs](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#extended-thinking-with-tool-use), thinking blocks **must** be echoed back with their `signature` during tool-use turns; OMA dropping signatures means `thinking: { type: 'enabled' }` users immediately break the protocol.
- **Gemini**: per [thought signatures docs](https://ai.google.dev/gemini-api/docs/thought-signatures), Gemini 2.5 returns thoughtSignature on function calls when thinking is enabled; **Gemini 3+ rejects with HTTP 400 if signatures are missing in subsequent turns**.
- **OpenAI**: incoming is fine; the gap is request-side (`reasoning_effort` not exposed for o1/o3/gpt-5).

Beyond protocol compliance, preserving reasoning across turns has measurable effect: re-derivation cost (each turn re-thinks from scratch) and quality regression on multi-step tasks.

## Verified current state (file:line refs)

**Anthropic** (`src/llm/anthropic.ts`)
- `fromAnthropicContentBlock:160-193` extracts thinking text but drops `signature` and falls through on `redacted_thinking` (becomes `[unsupported block type: redacted_thinking]`).
- `toAnthropicMessages:127-134` filters out reasoning via `SerializableContentBlock`.
- Streaming `thinking_delta:335-337` emits text deltas but no `signature_delta` collection.
[...]

PR #205, opened 2026-05-03
feat(llm): preserve reasoning blocks across Anthropic and Gemini turns (#200)

MyPrototypeWhat:
## Summary

Phase 1 of #200 — close the protocol gap where multi-turn extended thinking silently drops the signatures Anthropic and Gemini 3 require to continue a conversation. Scope follows the [owner's approval comment](https://github.com/JackChen-me/open-multi-agent/issues/200#issuecomment) verbatim.

### IR changes (`src/types.ts`)

- `ReasoningBlock.signature` — Anthropic `thinking.signature` and Gemini 3 thought-summary `thoughtSignature`. Required to echo back unchanged on Anthropic; recommended on Gemini 3.
- `ReasoningBlock.redactedData` — Anthropic `redacted_thinking.data`. Block must still be echoed back on subsequent turns.
- `ToolUseBlock.signature` — Gemini `thoughtSignature` for functionCall Parts. Required by Gemini 3+ ("required for function calling signatures" per Gemini docs).
- `AgentConfig.thinking: { enabled, budgetTokens?, effort? }` — explicit config. `effort` carried separately from `budgetTokens` per the owner's open-question resolution.

### Adapter changes

**`src/llm/anthropic.ts`** — extract `signature` from `thinking` blocks; new `redacted_thinking` branch on incoming; outgoing `toAnthropicContentBlockParam` handles reasoning blocks (`signature` → `thinking` block param, `redactedData` → `redacted_thinking` block param, neither → drop). New `toAnthropicThinkingParam` helper forwards `thinking: { type: 'enabled', budget_tokens }` on both `chat()` and `stream()`. Default budget 1024 (SDK comment: ``Must be ≥1024 and less than max_tokens``).

**`src/llm/gemini.ts`** — `thoughtSignature` is a top-level `Part` field per the `@google/genai` `Part` schema (verified at `genai.d.ts:8675`), NOT nested under `functionCall`. Extracted on both functionCall and thought-summary text Parts (Gemini 3 may sign both); echoed back on outgoing when present, dropped when unsigned to keep context lean. New `toGeminiThinkingConfig` helper forwards `thinkingConfig: { includeThoughts, thinkingBudget }`. `fromGeminiPart` helper extracted so `chat()` and `stream()` share one parsing implementation.

**`src/llm/bedrock.ts`** — comment refresh only. IR now carries the signature fields the previous comment said were missing; Bedrock-side wire-up to `reasoningContent.reasoningText.signature` is tracked as a follow-up.

**`src/agent/agent.ts` + `src/agent/runner.ts`** — one-line wire-ups to thread `thinking` from `AgentConfig` through `RunnerOptions` into `LLMChatOptions`.

### Tests

+23 across `tests/anthropic-adapter.test.ts` (+12) and `tests/gemini-adapter-contract.test.ts` (+11):

- Incoming extraction (signature, redactedData, thoughtSignature on both functionCall and thought-summary parts)
- Outgoing round-trip echo for all signed blocks
- Request-param forwarding (default 1024 for Anthropic, includeThoughts default for Gemini, omitted when disabled/absent)
- Negative path: unsigned reasoning dropped on outgoing for both adapters

`756/756` passing (baseline `733` + `23` new). `npm run lint` clean for this PR (3 pre-existing `bedrock.ts` errors are from a missing optional peer dep `@aws-sdk/client-bedrock-runtime`, unrelated and present on `main`).

### Cross-provider asymmetry to flag

Important detail captured in the IR JSDoc: the same `signature` field name carries different protocol semantics depending on origin — Anthropic signs `ReasoningBlock`s (required round-trip), Gemini signs `Part`s containing functionCalls (required) or text+`thought:true` summaries (recommended). The IR doesn't try to model the difference; the adapters know how to interpret their own `signature`s.

### Deferred to follow-up PRs (per owner)

- OpenAI `reasoning_effort` forwarding (explicit `effort` on `ThinkingConfig`, owner asked to keep separate from this PR).
- Anthropic adaptive thinking (Claude 4.7+ accepts only `{type: 'adaptive'}` and rejects `enabled` with HTTP 400). JSDoc on `toAnthropicThinkingParam` documents the constraint.
- Anthropic 4.5 `interleaved-thinking-2025-05-14` beta header detection.
- Bedrock signature wire-up (separate adapter rewrite).

Streaming `signature_delta` accumulation is intentionally NOT separate work — the existing `stream.finalMessage()` path returns the assembled SDK content with signatures intact, so the new `fromAnthropicContentBlock` extraction handles streaming "for free" (covered by the `extracts signature from streamed final message` test).

## Test plan

- [x] `npm test` — 756/756 (baseline 733 + 23 new)
- [x] `npm run lint` — no new errors (3 pre-existing bedrock-related errors confirmed via stash)
- [x] Anthropic implementation cross-checked against `platform.claude.com/docs/en/docs/build-with-claude/extended-thinking` and SDK type definitions in `@anthropic-ai/sdk@0.52.0`
- [x] Gemini implementation cross-checked against `ai.google.dev/gemini-api/docs/thinking` and `@google/genai` `Part` schema (verified `thoughtSignature` is top-level on Part, not nested)


JackChen-me:
`toAnthropicThinkingParam` defaults `budget_tokens` to 1024 without knowing `maxTokens`. Anthropic requires `budget_tokens < max_tokens`, so `thinking: { enabled: true }` with `maxTokens <= 1024`, or any `budgetTokens >= maxTokens`, is a guaranteed 400 at runtime. Pass `maxTokens` in and clamp, or validate and throw before the request.

Rest looks right for the phase-1 scope.

MyPrototypeWhat:
Good catch, you're right. Verified against the docs ("`budget_tokens` must be set to a value less than `max_tokens`") — the interleaved-thinking exception doesn't apply since the beta header isn't wired up in this PR.

Pushed `a0c721c` taking the validate-and-throw path:
- `budget < 1024` → throws (SDK-documented minimum)
- `budget >= maxTokens` → throws with a message identifying which constraint failed and how to fix it
- Threaded `effectiveMaxTokens = options.maxTokens ?? 4096` into both `chat()` and `stream()` so the validator sees the same value the request uses
- +3 tests including a regression for the exact scenario you flagged (default budget 1024 + `maxTokens` 1024)

Went with throw rather than clamp because OMA's pattern is explicit-over-magic and silently changing a user-specified `budgetTokens` would surprise debuggers. Happy to switch to clamp if you'd prefer.

`759/759` still green.

Issue #25, opened 2026-04-02
[Feature] OpenAI-compatible provider integrations (DeepSeek, Mistral, Groq, Zhipu, MiniMax, Qwen, Moonshot, Doubao)


JackChen-me:
These providers offer OpenAI-compatible APIs and should work via existing `openai` provider with custom `baseURL`. Each one needs:

1. **Verify** tool-calling works correctly
2. **Add a working example** to `examples/`
3. **Document** in README under Supported Models

| Provider | baseURL | Env var |
|----------|---------|--------|
| DeepSeek | `https://api.deepseek.com` | `DEEPSEEK_API_KEY` |
| Mistral | `https://api.mistral.ai/v1` | `MISTRAL_API_KEY` |
| Groq | `https://api.groq.com/openai/v1` | `GROQ_API_KEY` |
| Zhipu GLM | `https://open.bigmodel.cn/api/paas/v4` | `ZHIPU_API_KEY` |
| MiniMax | `https://api.minimax.chat/v1` | `MINIMAX_API_KEY` |
| Qwen (DashScope) | `https://dashscope.aliyuncs.com/compatible-mode/v1` | `DASHSCOPE_API_KEY` |
| Moonshot (Kimi) | `https://api.moonshot.cn/v1` | `MOONSHOT_API_KEY` |
| Doubao (ByteDance) | `https://ark.cn-beijing.volces.com/api/v3` | `ARK_API_KEY` |

Contributions for any single provider are welcome — no need to cover all at once.

JackChen-me:
## How to contribute

Pick a provider, test it, submit a PR. One provider per PR.

### Status (updated 2026-05-28)

Original 8 providers all shipped.

| Provider | Status | Contributor |
|----------|--------|-------------|
| DeepSeek | Done, first-class `provider: 'deepseek'` | hkalex (#114) |
| MiniMax | Done, first-class `provider: 'minimax'` | hkalex (#113) |
| Groq | Done, `openai` + `baseURL`, example 19 | mvanhorn (#121) |
| Mistral | Done, `openai` + `baseURL`, `examples/providers/mistral.ts` | mvanhorn (#206) |
| Zhipu GLM | Done, `openai` + `baseURL`, `examples/providers/zhipu.ts` | nuthalapativarun (#231) |
| Qwen (DashScope) | Done, `openai` + `baseURL`, `examples/providers/qwen.ts` | goodneamtakenbydogs (#257) |
[...]

PR #206, opened 2026-05-03
feat: add Mistral provider example and README entry

mvanhorn:
Refs #25.

Adds Mistral via the existing OpenAI-compatible adapter, following the same shape as the merged Groq, MiniMax, Grok, and DeepSeek precedents.

Files

- examples/providers/mistral.ts — three-agent (architect / developer / reviewer) team that builds a small Express.js REST API. Uses provider: 'openai' + baseURL: 'https://api.mistral.ai/v1' with apiKey from MISTRAL_API_KEY. Default model mistral-large-latest, with mistral-medium-latest documented as an alternative in the file header.
- README.md — adds a Mistral row to the OpenAI-compatible providers table next to Groq; appends "Mistral example" to my contributor line; trims Mistral from the "all plug in the same way" sentence since it now has an example.

Verification

- npm run lint (tsc --noEmit) passes locally.
- The example does not run during build; it requires MISTRAL_API_KEY to execute.

Issue table coverage after this PR: Groq, MiniMax, Grok, DeepSeek, Mistral done. Still open in #25: Zhipu GLM, Qwen, Moonshot, Doubao.

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,16 +355,17 @@ The framework ships a wired-in provider name for each of these. You set `provide
 
 No bundled shortcut, but it works the same. Use `provider: 'openai'` and point `baseURL` at any server that speaks OpenAI Chat Completions.
 
-| Service | Config | Env var | Example model |
-|---------|--------|---------|---------------|
-| Ollama (local) | `provider: 'openai'` + `baseURL: 'http://localhost:11434/v1'` | none | `llama3.1` |
-| vLLM (local) | `provider: 'openai'` + `baseURL` | none | (server-loaded) |
-| LM Studio (local) | `provider: 'openai'` + `baseURL` | none | (server-loaded) |
-| llama.cpp server (local) | `provider: 'openai'` + `baseURL` | none | (server-loaded) |
-| OpenRouter | `provider: 'openai'` + `baseURL: 'https://openrouter.ai/api/v1'` + `apiKey` | `OPENROUTER_API_KEY` | `openai/gpt-4o-mini` |
-| Groq | `provider: 'openai'` + `baseURL: 'https://api.groq.com/openai/v1'` | `GROQ_API_KEY` | `llama-3.3-70b-versatile` |
+| Service | Config | Env var | Example model | Notes |
+|---------|--------|---------|---------------|-------|
+| Ollama (local) | `provider: 'openai'` + `baseURL: 'http://localhost:11434/v1'` | none | `llama3.1` | |
+| vLLM (local) | `provider: 'openai'` + `baseURL` | none | (server-loaded) | |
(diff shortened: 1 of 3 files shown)


JackChen-me:
`runTeam()` creates a coordinator agent from the orchestrator defaults, and that coordinator does not inherit `baseURL` / `apiKey` from the team agents. Since this example only sets `defaultProvider: 'openai'`, the coordinator will fall back to `api.openai.com` (or fail if `OPENAI_API_KEY` is unset) instead of using Mistral.

Please add:

```ts
defaultBaseURL: 'https://api.mistral.ai/v1',
defaultApiKey: process.env.MISTRAL_API_KEY,
```

to the orchestrator constructor.

CodingBangboo:
Thanks for the contribution @mvanhorn 

Two more things worth flagging in addition to the coordinator fix:
**1. Add an upfront `MISTRAL_API_KEY` guard.** If the env var is unset, `apiKey: process.env.MISTRAL_API_KEY` becomes `undefined` and the OpenAI SDK falls back to `OPENAI_API_KEY` : which then gets sent to `api.mistral.ai` and returns a confusing 401. Once `defaultApiKey` is added per @JackChen-me's earlier comment, the same fallback would let the coordinator silently route to OpenAI (and bill that account) when `OPENAI_API_KEY` happens to be set. `examples/providers/openrouter.ts` handles this with a simple guard at the top:

​```
if (!process.env.MISTRAL_API_KEY) {
  throw new Error('MISTRAL_API_KEY environment variable must be set.')
}
​```

**2. Same coordinator bug exists in `examples/providers/groq.ts`.** It uses the same `provider: 'openai'` + custom `baseURL` shape and also omits `defaultBaseURL` / `defaultApiKey`, so its coordinator falls back to `api.openai.com` too. Probably worth fixing in this PR (or a quick follow-up) along with a `GROQ_API_KEY` guard or might be better in another small pr.

mvanhorn:
Thanks @CodingBangboo, both flagged in d7f24d3:

1. Added the MISTRAL_API_KEY guard at the top of mistral.ts, matching the openrouter.ts pattern. Also pulled the URL and key into MISTRAL_BASE_URL / MISTRAL_API_KEY constants so the orchestrator config can reuse them.

2. Same fix applied to groq.ts in the same commit (GROQ_API_KEY guard + defaultBaseURL / defaultApiKey on the orchestrator). Easier to land together than as a follow-up.

Also wired up @JackChen-me's earlier ask: defaultBaseURL and defaultApiKey are now passed to OpenMultiAgent so the coordinator inherits the provider settings instead of falling back to api.openai.com.

CodingBangboo (approved):
lgtm! Will wait for @JackChen-me's final check and then good to merge.

PR #213, opened 2026-05-07
feat(memory): add turn-count TTL to SharedMemory entries

MyPrototypeWhat:
## Summary

Adds optional turn-count expiry to shared-memory entries so agents can say *"remember this for the next 3 turns then forget"* without manual cleanup. Common need in multi-agent runs (e.g. ephemeral hand-offs, intermediate scratch values that shouldn't pollute later prompts).

Ported from [`context-chef`](https://github.com/MyPrototypeWhat/context-chef)'s `Memory.set({ ttl })` / `advanceTurn()` design — same algorithm (`expiresAtTurn = currentTurn + ttl`, lazy filter on read), narrowed to turn-count only (no ms-TTL — chef has it, but it's an edge case with no clear OMA caller; can add later if asked).

## API additions (all backward compatible)

| Surface | What | Notes |
|---|---|---|
| `MemoryEntry.expiresAtTurn?: number` | optional field | only set by stores that support TTL |
| `MemoryStore.setWithExpiry?(key, value, expiresAtTurn, metadata?)` | optional method | stores that don't implement it silently lose TTL — `writeExpiring` degrades to `set` |
| `SharedMemory.advanceTurn()` | increments internal counter | called by orchestrator (see below) |
| `SharedMemory.getTurnCount()` | reads counter | observability / tests |
| `SharedMemory.writeExpiring(agent, key, value, ttlTurns, metadata?)` | parallel to `write` | tags entry with computed `expiresAtTurn` |

`SharedMemory.read` / `listAll` / `listByAgent` / `getSummary` all filter expired entries and delete them as a side effect (lazy cleanup, no background timer).

## Orchestrator integration

`runTeam` and `runTasks` advance the shared memory turn counter once per completed task — right after the task result is written to shared memory (orchestrator.ts:711). Standalone `runAgent` does **not** advance — there's no team turn boundary in single-agent runs.

## Backward compatibility

- Existing `SharedMemory.write(...)` and `MemoryStore.set(...)` callers: zero change. All 5 new fields/methods are optional.
- Custom `MemoryStore` implementations (Redis, SQLite, etc.): zero change required. They opt into TTL by implementing `setWithExpiry`; otherwise `writeExpiring` falls back to `set` and the entry never expires.

## Tests (+5)

- `writeExpiring` entries readable until counter reaches expiry (`currentTurn >= expiresAtTurn`)
- `write` entries persist regardless of turn count
- `listAll` / `listByAgent` / `getSummary` all filter expired
- `writeExpiring` degrades to plain `set` when store lacks `setWithExpiry`
- `getTurnCount` reflects `advanceTurn` calls

`788/788` passing. `npm run lint` clean.

## Design alternatives considered

- **Single `write` with options object** — would be cleaner long-term but breaks the existing 4-positional `write(agent, key, value, metadata?)` signature. Parallel `writeExpiring` keeps the surface additive. Happy to consolidate later if you'd prefer one method with options — let me know.
- **Magic key in `metadata`** — avoids interface changes entirely but is a "convention not contract" pattern that's easy to misuse. Optional method on the interface is more typed.
- **ms-based TTL** — chef supports both turn-based and ms-based expiry. Not ported — turn-based covers the common multi-agent hand-off case, and ms-TTL would need a wall-clock semantics decision. Easy follow-up if requested.

## Test plan

- [x] `npm test` — 788/788 (baseline 783 + 5 new)
- [x] `npm run lint` — clean
- [x] Verified backward compat: existing `tests/shared-memory.test.ts` still passes unchanged
- [x] `MemoryStore` interface change is purely additive — `tsc` accepts existing custom-store impls

### DIFF
--- a/src/memory/shared.ts
+++ b/src/memory/shared.ts
@@ -145,8 +145,10 @@ export class SharedMemory {
    * the optional method.
    *
    * @param ttlTurns - Number of turns the entry should remain readable for.
-   *                   Must be a positive integer; `0` means the entry is
-   *                   already expired and won't be returned by reads.
+   *                   Must be an integer ≥ 1; throws {@link RangeError}
+   *                   otherwise.
+   *
+   * @throws {RangeError} when `ttlTurns` is not an integer or is less than 1.
    *
    * @remarks
    * In parallel batch execution (`runTasks` / `runTeam` running multiple
@@ -164,6 +166,12 @@ export class SharedMemory {
     ttlTurns: number,
     metadata?: Record<string, unknown>,
   ): Promise<void> {
+    if (!Number.isInteger(ttlTurns) || ttlTurns < 1) {
+      throw new RangeError(
+        `SharedMemory.writeExpiring: ttlTurns must be an integer ≥ 1 (got ${ttlTurns}). ` +
+          'Use write() for entries that should never expire.',
+      )
+    }
     const namespacedKey = SharedMemory.namespaceKey(agentName, key)
     const fullMetadata = { ...metadata, agent: agentName }
     if (typeof this.store.setWithExpiry === 'function') {
@@ -184,15 +192,15 @@ export class SharedMemory {
    *
    * Returns `null` when the key is absent **or** when the entry has expired
    * (per its `expiresAtTurn` against the current turn counter). Expired
-   * entries are deleted from the underlying store as a side effect.
+   * entries are filtered out but **not** deleted from the underlying store —
+   * deletion is left to the store impl (Redis has native TTL, Postgres a
+   * cron, etc.). Reading is therefore safe to call from concurrent processes
+   * without the risk of stomping on a fresh write to the same key.
    */
   async read(key: string): Promise<MemoryEntry | null> {
     const entry = await this.store.get(key)
     if (entry === null) return null
-    if (this.isExpired(entry)) {
-      await this.store.delete(key)
-      return null
-    }
+    if (this.isExpired(entry)) return null
     return entry
   }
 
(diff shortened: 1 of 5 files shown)


JackChen-me:
Lazy cleanup can delete a fresh rewrite in custom stores.

`read()` / `filterExpired()` decide expiry from a `get()`/`list()` snapshot, then `delete(key)`. If a Redis/Postgres-backed store writes a new value for the same key in between, cleanup removes the fresh value too. I'd make this non-destructive or delete only when the current `expiresAtTurn` still matches.

CodingBangboo (commented):
Hi @MyPrototypeWhat thanks for the contribute. Just a very quick comment, would you think that `ttlTurns` might accept <= 0 values silently? Should we add a guard to make the API stronger?
```
if (!Number.isInteger(ttlTurns) || ttlTurns < 1) throw new RangeError(...)
```

MyPrototypeWhat:
Good catch — fixed in `c31a3eb`.

Went with the non-destructive option: `read()` / `listAll()` / `listByAgent()` / `getSummary()` now just filter expired entries from the returned view; nothing in `SharedMemory` calls `store.delete()` for TTL purposes. Stores own their own retention (Redis has native EXPIRE, Postgres a cron, in-memory accepts unbounded growth in pathological cases — acceptable for typical agent-run scale).

Didn't go with the compare-and-swap `delete-only-when-expiresAtTurn-still-matches` route because true CAS would need a new optional store method, and the value of opportunistic cleanup-on-read isn't worth that surface area when most real backends have native TTL anyway.

New regression test asserts the underlying store still contains the expired entry after `SharedMemory` filters it out, so this won't quietly slip back.

MyPrototypeWhat:
Thanks @CodingBangboo, agreed — fixed in `c31a3eb`. The previous JSDoc was self-contradictory ("Must be a positive integer" vs "`0` means already expired"). Now `writeExpiring` throws `RangeError` when `!Number.isInteger(ttlTurns) || ttlTurns < 1`, so `0` / negatives / `NaN` / `Infinity` / non-integers all fail loudly instead of silently going dead-on-arrival. Test covers all five cases.

JackChen-me (approved):
LGTM. Both review comments are addressed and targeted tests pass.

PR #215, opened 2026-05-07
[Bug] Fix Truncation and Tighten Coordinator's Dependency Guidance

### LINKED ISSUE #208: [Discussion/Behavior Design] Coordinator adds dependencies beyond each agent's declared input contract
# Coordinator adds dependencies beyond each agent's declared input contract

## Summary

`runTeam`'s coordinator silently widens an agent's `dependsOn` set beyond what the agent's system prompt declares as its input. When a system prompt names a single input ("You consume the architectural blueprint."), the coordinator wires multiple parents anyway. When the prompt is exhaustive ("Both inputs are required."), the coordinator obeys exactly. The inconsistency means roster authors can't predict the DAG from the prompts they wrote, and the surprise costs both parallelism and tokens — every extra parent injects that parent's full output into the dependent task's prompt (orchestrator.ts:800-811).

## Reproducer

`plan-only-dag_stress_test.ts` at https://github.com/CodingBangboo/open-multi-agent/blob/planOnly_with_examples/examples/basics/plan-only-dag_stress_test.ts
, run with `planOnly: true`. 7-agent roster, Sonnet 4.6 coordinator, no coordinator overrides.

## Observed behaviour

| Agent | Inputs declared in system prompt | Coordinator wired | Faithful? |
|---|---|---|---|
| requirements-analyst | (none) | (root) | yes |
| security-auditor | requirements | requirements | yes |
| architect | requirements | requirements + **security** | **no** |
| threat-modeler | "the architectural blueprint with the security requirements addendum. Both inputs are required." | architect + security | yes |
| frontend-planner | "the architectural blueprint" | architect + **security + threat-model** | **no** |
| backend-planner | "the architectural blueprint" | architect + **security + threat-model** | **no** |
| project-manager | (enumerates all six artefacts) | all 6 | yes |

Result: 17 edges instead of the 12 the roster's input contracts imply.

## Why this matters

1. **Parallelism loss.** With the actual DAG, every level past requirements serialises — the longest path grows from 4 levels to 6. With explicit input contracts honoured, frontend/backend can run in parallel with threat-modeler instead of after it.
2. **Token cost.** Per `buildTaskPrompt` (orchestrator.ts:800-811), the default behaviour is "default-deny": each parent's full output is concatenated into the child's prompt. Extra parents = extra input tokens.
3. **Predictability.** Roster authors who carefully scope each agent's inputs can't anticipate the resulting DAG. The same goal + roster will not produce the same DAG run-to-run, and the spread is governed by coordinator judgement, not the user's design.

## Why I think this is a real defect, not just a tuning question

The coordinator already faithfully translates *exhaustive* input declarations (threat-modeler, project-manager). The defect is that when the input list is short, the coordinator volunteers extras instead of treating brevity as authoritative. Either both behaviours are correct and the user has no way to say "no, that's the full list," or the exhaustive case is correct and the short case is a regression.

## Proposed fixes

1. **Coordinator prompt should treat per-agent input declarations as authoritative.** Add a line to `buildCoordinatorOutputFormatSection` (orchestrator.ts:1483-1497): "When wiring `dependsOn`, prefer the minimum set of upstream tasks needed for the assignee to satisfy its system prompt. Do not add a dependency unless the assignee's system prompt indicates it needs that input."
2. **Stop truncating system prompts in the roster.** orchestrator.ts:1473 slices each prompt to 120 chars. For longer prompts the trailing input clause may be clipped, weakening the contract the coordinator sees. Either remove the slice or raise it to a length that preserves typical input-declaration sentences.
3. **Optional second mode**: a coordinator preset for "respect declared inputs strictly" vs the current "infer extras," exposed via `coordinator.instructions` so existing users who depend on the broader behaviour aren't broken silently.

## Environment

- Coordinator model: `claude-sonnet-4-6` (Anthropic)
- Branch: `planOnly_with_examples` from forked repo
- Coordinator tokens for the run: input 2365, output 1350

Open to discussion.

CodingBangboo:
## What

Two changes in `src/orchestrator/orchestrator.ts`, both in `buildCoordinatorOutputFormatSection` / roster rendering:

1. **`orchestrator.ts:1473`** — removed the `slice(0, 120)` on each agent's `systemPrompt` so the coordinator sees the full input-contract sentence.
2. **`orchestrator.ts:1492-1500`** — added a "Dependency Rules" block. It **tightening coordinator guidance toward the minimum dependency set**, but **NO** enforceable contract by default.

The prompt change stays on the approved side of our discussion:
- The new "Dependency Guidance" block is phrased as preferences ("prefer the minimum set", "lean toward", "avoid", "when uncertain, prefer fewer"), not hard rules. The coordinator can still wire an extra parent if it judges the assignee genuinely needs it to match the current `runTeam()` behavior.
- There is no post-validation of the returned DAG against declared inputs. Nothing rejects or rewrites the coordinator's output.
- Default `runTeam()` behavior does shift a little — the coordinator will tend to wire fewer deps than before — but that softening is the explicitly approved goal, not strict enforcement. 

After all, this prompt change will only nudge the LLM behavior a little, no hard enforcement. **We can remove this change if it is still out of scope**

Tests:

- `tests/coordinator-dependency-contract.test.ts` — mocked-LLM regression guard that the framework forwards coordinator `dependsOn` faithfully (no widening or silent drops at the framework layer).
- `tests/e2e/coordinator-dependency-widening.test.ts` — `RUN_E2E=1`-gated test using the issue's 7-agent roster verbatim, asserting `frontend-planner` / `backend-planner` / `architect` don't gain extra parents.

## Why
fixes #208 

## Out of Scope (should not be in this PR per discussion in #208)
No opt-in strict mode / preset 
- No `coordinator.instructions` strict-mode preset.
- No change to default `runTeam` semantics (No strict declared-input-as-contract by default)

## Limitations

1. **Coordinator input tokens grow.** Removing the 120-char slice ships full system prompts into the coordinator call. Net token impact depends on roster size vs. parallelism win.
2. **E2E coverage is opt-in.** The real-LLM regression test is skipped unless `RUN_E2E=1` + `ANTHROPIC_API_KEY`. Meanwhile I only ensured that the E2E test works fine with the sonnet model. For other LLM's behavior, no strict guarantee yet. 
3. **Unit test only proves a weaker claim.** The framework-level forwarding test passes today even without the prompt change; it guards against a *different* class of regression (the framework editing the DAG) than the one the issue reports (LLM judgment). 

## Open Question

Should the roster cap be removed entirely or raised to a numeric limit (e.g. 800–1000 chars)? In this PR I removed the cap for simplicity.


JackChen-me:
LGTM.

Issue #30, opened 2026-04-02
[P0] Task Retry — maxRetries with exponential backoff


JackChen-me:
## Context

LLM API calls are inherently unreliable — rate limits, transient 500s, network hiccups. Currently, a single failed LLM call fails the entire task, which cascades to all dependent tasks. For production workflows this is unacceptable.

Task-level retry is the simplest reliability improvement with the highest impact.

## Proposed Implementation

Add optional retry fields to task config:

```typescript
const tasks = [
  {
    title: 'Generate report',
    assignee: 'analyst',
    maxRetries: 2,           // default: 0 (no retry)
    retryDelayMs: 1000,      // default: 1000
    retryBackoff: 2,         // default: 2 (exponential)
    description: '...',
  },
]
```

When a task fails and `maxRetries > 0`:
1. Wait `retryDelayMs * retryBackoff^attempt` milliseconds
2. Re-run the task with a fresh agent conversation (no corrupted state)
3. Emit `task_retry` event via `onProgress` with attempt number and error
4. After all retries exhausted, cascade failure as today

### Key decisions
- Retry at task level (not LLM call level) — simpler, covers all failure modes
- Fresh conversation per retry — avoids poisoned context from failed attempts
- Exponential backoff — respects rate limits
- `onProgress` event for observability — callers can log/alert on retries

## Acceptance Criteria

- [ ] `maxRetries`, `retryDelayMs`, `retryBackoff` on task config
- [ ] Exponential backoff between retries
- [ ] Fresh agent state per retry (no conversation carry-over)
- [ ] `task_retry` event emitted via `onProgress`
- [ ] Cascade failure only after all retries exhausted
- [ ] Works with both `runTeam()` and `runTasks()`
- [ ] Tests cover: retry succeeds on 2nd attempt, all retries exhausted, backoff timing

JackChen-me:
**Source:** Reddit community feedback
- u/Specialist_Golf8133

PR #224, opened 2026-05-13
feat(examples): add mcp-notte integration example

sam-notte:
## What

Adds `examples/integrations/mcp-notte.ts` mirroring the shape of `mcp-github.ts`. Wires Notte (notte.cc, hosted browser infra) as a tool source via the existing `connectMCPTools()` path you shipped in #89. No core code changed.

## Why

The repo has six built-in tools and two example MCP integrations (`mcp-github.ts`, `with-engram/`) but no example for browser or web-research. Many real agent tasks need to drive a live browser: navigate, fill forms, scrape JS-rendered pages, get past anti-bot walls. Notte ships an MCP server (`notte-mcp`) over stdio, so this is the minimal wire-up.

## Context

- Notte recently merged into [Bitterbot-AI/bitterbot-desktop](https://github.com/Bitterbot-AI/bitterbot-desktop) as a bundled skill ([#30](https://github.com/Bitterbot-AI/bitterbot-desktop/pull/30), 2026-05-05).
- Notte is hosted browser infra: residential proxies, captcha solving, session vault, stealth profiles, and deploy-as-Function (a one-click way to turn an automation into a scheduled API endpoint). Open source: [github.com/nottelabs/notte](https://github.com/nottelabs/notte) (1,953 stars, SSPL-1.0).

## Test plan

- [ ] `NOTTE_API_KEY=xxx GEMINI_API_KEY=xxx npx tsx examples/integrations/mcp-notte.ts`
- [ ] Expected: agent returns 5 HN posts as JSON
- [ ] Verified the file follows the same shape as `mcp-github.ts` (single-Agent pattern, `try { ... } finally { await disconnect() }` cleanup)

## Licence

MIT, same as repo.

---

*If you'd prefer this routed via the `customTools` pattern (per the `with-engram/team-research.ts` shape) instead of single-Agent, happy to refactor; the single-Agent shape felt closer to the `mcp-github.ts` precedent.*

### DIFF
--- a/examples/integrations/mcp-notte.ts
+++ b/examples/integrations/mcp-notte.ts
@@ -1,0 +1,67 @@
+/**
+ * MCP Notte Tools
+ *
+ * Connect Notte's MCP server (notte-mcp) over stdio and register all exposed
+ * Notte tools — observe, click, fill, scrape, navigate, run agent — as
+ * standard open-multi-agent tools. Gives any agent a hosted browser with
+ * residential proxies, captcha solving, stealth sessions, and credential
+ * vault for authenticated sites.
+ *
+ * Notte (notte.cc) is hosted browser infrastructure for AI agents:
+ *   https://github.com/nottelabs/notte
+ *
+ * Run:
+ *   npx tsx examples/integrations/mcp-notte.ts
+ *
+ * Prerequisites:
+ *   - NOTTE_API_KEY  (get one at https://notte.cc/dashboard)
+ *   - GEMINI_API_KEY (or set AGENT_PROVIDER + the matching key)
+ *   - uv installed: `brew install uv`  (or `pipx install notte-mcp` and use `notte-mcp` directly)
+ *   - @modelcontextprotocol/sdk installed
+ */
+
+import { Agent, ToolExecutor, ToolRegistry, registerBuiltInTools } from '../../src/index.js'
+import { connectMCPTools } from '../../src/mcp.js'
+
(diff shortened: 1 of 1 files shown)


CodingBangboo (changes requested):
Thanks for the contribution @sam-notte! I wouldn't say the system prompt is perfect for now because there is no built-in tool that does http/fetch specifically. Currently only {bash, file_read, file_write, file_edit, grep, glob} are registered. Would it be better to address it like "prefer Notte over bash-driven curl", or just drop the line because pointing at a phantom built-in is usually not desired. 

In addition, "residential proxies, captcha solving, session vault, stealth profiles, deploy-as-Function" are not in this example PR. I would trim the marketing language a bit.

Also, for this PR, note this is a `vendor-contribution` @JackChen-me. Please also note the uvx toolchain dep mentioned in the PR.

JackChen-me:
I agree with @CodingBangboo here. This is a vendor-specific integration for a third-party hosted browser service, and OMA already supports this path through the generic `connectMCPTools()` API.

I don’t think we should maintain the full Notte example in this repo. Please move the “Using Notte MCP with open-multi-agent” guide to the Notte docs/repo; we can consider linking to it from OMA docs.

Also agreed on the current issues: full `process.env` is passed to a third-party MCP subprocess, the header claims `AGENT_PROVIDER` support while the code hardcodes Gemini, and the example reads too much like product positioning.

PR #225, opened 2026-05-14
Add Moonshot provider example

mzl2233:
Adds a Moonshot (Kimi) example that uses the existing OpenAI-compatible adapter with `MOONSHOT_API_KEY` and the official base URL. Also updates the provider docs and README supported-provider summary so Moonshot is discoverable alongside the other OpenAI-compatible services.

Tested with `npm run lint` and `npm test`.


JackChen-me:
Can we mirror the MiniMax setup here? It supports a base URL env override and documents the regional endpoint choice, which fits Kimi too since the English docs use `.ai` and the CN docs use `.cn`.

JackChen-me:
Still picking this up, or should I free up the slot in #25?

JackChen-me:
Closing as stale to free the #25 slot.

PR #226, opened 2026-05-14
Add Moonshot provider example

mzl2233:
Adds a Moonshot AI (Kimi) provider example using the OpenAI-compatible adapter, including API key validation, base URL configuration, and the standard three-agent REST API workflow used by the other provider samples. Also updates the supported-provider list in the README so users can discover the new example.


JackChen-me:
duplicate #225

Issue #223, opened 2026-05-13
[Proposal] Cross-provider reasoning fallback (`<thinking>` text downgrade)


MyPrototypeWhat:
> Follow-on to #200 (closed as completed). #200 closed the **within-provider** protocol-correctness gap (Anthropic / Gemini signature round-trip via #205, OpenAI `reasoning_effort` forwarding via #209). This proposes a separately-scoped fix for the orthogonal **cross-provider** silent-drop that #200 didn't cover.

## Motivation

After #205 + #209, OMA correctly preserves `ReasoningBlock`s when the producing and consuming adapter are the **same** provider. There's still a silent drop whenever a `ReasoningBlock` produced by provider A ends up replayed to provider B that can't natively echo it. The block disappears without any signal — no warning, no metadata.

This bites in two real scenarios on `main`:

1. **`Agent.prompt()` mid-conversation model swap** (`agent/agent.ts:212-235`) — `prompt()` replays the persistent history to whichever adapter the new `provider`/`model` resolves to. Swapping from Anthropic to OpenAI loses the prior reasoning silently.
2. **Bedrock multi-turn**, both same-provider and cross-provider, until the existing Phase-1-style follow-up (`bedrock.ts:112-119` TODO) lands.

## Verified current state (file:line refs)

| Source → target | Behaviour today |
|---|---|
| Anthropic → Anthropic | Round-trips with `signature` ✅ (#205) |
| Gemini → Gemini | Round-trips with `signature` (carrying Gemini's `thoughtSignature`) ✅ (#205) |
| Anthropic → OpenAI / Azure / Copilot / Grok / DeepSeek / Qiniu | `toOpenAIAssistantMessage` (`openai-common.ts:166-196`) loops over `tool_use` / `text` only — `reasoning` blocks fall through and are dropped silently |
| Anthropic / OpenAI → Gemini | `toGeminiContents` (`gemini.ts:78-110`) explicitly drops unsigned reasoning, including cross-provider blocks (per its own JSDoc) |
| anything → Bedrock | `bedrock.ts:112-119` drops on input (TODO) |
| OpenAI / cross-provider → Anthropic | Even if we kept the text, sending it back as a structured `thinking` block without a valid Anthropic signature would fail the API's signature check — so structured re-promotion is **never** an option |

## Proposed design

[...]

PR #234, opened 2026-05-17
openai: downgrade reasoning blocks to <thinking> text fallback

matthewYang08:
## What

This PR adds a fallback for OpenAI wire conversion when assistant messages contain `reasoning` blocks in framework IR.

In `toOpenAIAssistantMessage`, `reasoning` blocks are now downgraded to inline `<thinking>...</thinking>` text and prepended to the next assistant text block. If no text block exists, the fallback text is still emitted as assistant content. For redacted reasoning (`redactedData`), the fallback emits `<thinking>[redacted]</thinking>`.

## Why

Cross-provider or non-native reasoning content could previously be dropped silently when converting assistant IR messages for OpenAI-compatible payloads. This keeps reasoning context visible instead of disappearing.

Refs #223

## Checklist

- [ ] `npm run lint` passes
- [x] `npm run test -- tests/openai-common.test.ts tests/openai-fallback.test.ts` passes

### DIFF
--- a/src/llm/openai-common.ts
+++ b/src/llm/openai-common.ts
@@ -72,6 +72,13 @@ export function getOpenAIReasoningText(source: unknown): string {
   return extractReasoningText((source as Record<string, unknown>)['reasoning_content'])
 }
 
+function reasoningBlockToThinkingText(block: ReasoningBlock): string {
+  if (typeof block.redactedData === 'string' && block.redactedData.length > 0) {
+    return '<thinking>[redacted]</thinking>'
+  }
+  return `<thinking>${block.text}</thinking>`
+}
+
 /**
  * Determine whether a framework message contains any `tool_result` content
  * blocks, which must be serialised as separate OpenAI `tool`-role messages.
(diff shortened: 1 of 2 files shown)


JackChen-me:
Withdrawing my earlier comment. Citing #223 as adopted spec was wrong.

The underlying problem is real, but this PR ships the conversion with no toggle at all. Any OpenAI-family response carrying `reasoning_content` can now become a `ReasoningBlock`, and that block is replayed as full `<thinking>` text on later OpenAI-family requests in the same history, even where the cross-provider case doesn't apply. This is a silent prompt-size/cost regression on upgrade with no opt-out and no truncation.

Needs an opt-in flag, off by default, before this can land.

matthewYang08:
> Withdrawing my earlier comment. Citing #223 as adopted spec was wrong.
> 
> The underlying problem is real, but this PR ships the conversion with no toggle at all. Any OpenAI-family response carrying `reasoning_content` can now become a `ReasoningBlock`, and that block is replayed as full `<thinking>` text on later OpenAI-family requests in the same history, even where the cross-provider case doesn't apply. This is a silent prompt-size/cost regression on upgrade with no opt-out and no truncation.
> 
> Needs an opt-in flag, off by default, before this can land.


Thanks, addressed your blocker on #234 with minimal scope:

- Reasoning replay is now behind explicit opt-in (`enableReasoningTextReplay` path), default remains off.
- Added bounded replay (`maxReasoningReplayChars`) with truncation.
- Prevented empty assistant message emission in reasoning-only/off path.
- Added/updated tests for default-off compatibility, opt-in behavior, truncation bound, and invalid max-value handling.

Could you please re-review when you have a moment?

JackChen-me:
LGTM

PR #247, opened 2026-05-25
docs(examples): add Bilig WorkPaper MCP integration

gregkonush:
## Summary

Adds a Bilig WorkPaper MCP integration example that follows the existing `mcp-github.ts` pattern: connect an MCP stdio server with `connectMCPTools()`, register the tools, and let an `Agent` use them.

The example pins `@bilig/workpaper@0.96.0`, initializes a demo pricing WorkPaper, asks the agent to read `Summary!B3`, write `Inputs!B3 = 0.4`, and verify the recalculated `Summary!B3` result.

## Validation

- `npm run build`
- `npm test -- tests/mcp-tools.test.ts`
- `AGENT_MODEL=gemini-2.5-flash npx tsx examples/integrations/mcp-bilig-workpaper.ts`
- `git diff --check`

The local agent run reported `Summary!B3` changed from `60000` to `96000` and that the workbook recalculated and persisted the edit.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,7 +2,7 @@
 
 Runnable scripts demonstrating `open-multi-agent`. Organized by category — pick one that matches what you're trying to do.
 
-All scripts run with `npx tsx examples/<category>/<name>.ts` and require the corresponding API key in your environment.
+All scripts run with `npx tsx examples/<category>/<name>.ts`. Scripts that call a model require the corresponding API key in your environment.
 
 ---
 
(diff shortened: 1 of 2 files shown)


JackChen-me:
Holding off on this one. `examples/integrations/` carries an implicit reciprocity bar: vendor lists open-multi-agent in their own integrations too. Precedent is `with-engram`, where Engram's README at https://github.com/Agentscreator/engram-memory has a dedicated OMA callout. Bilig's integrations list at https://proompteng.github.io/bilig/ doesn't currently include OMA.

Two things to fix before reopening: pin `@bilig/workpaper` to a specific version instead of `@latest` so the example is reproducible, and trim the assertion harness. `mcp-github.ts` is ~60 lines driving the MCP tools through an actual `Agent`, which is the pattern users want to learn from.

gregkonush:
Thanks, that makes sense. I pushed the cleanup in `2e16f33`.

- pinned the example to `@bilig/workpaper@0.96.0`
- trimmed it down to the existing `mcp-github.ts` pattern: `connectMCPTools()` -> `Agent` -> real MCP tool calls
- added the reciprocal Bilig integration page: https://proompteng.github.io/bilig/open-multi-agent-workpaper-mcp.html

Validation after the rebase:

```text
npm run lint
npm test
git diff --check
```

The Bilig page is live now and links back to this PR, so the reciprocity piece is covered from our side.

JackChen-me:
Reciprocity page looks solid, tech changes are clean, CI green. Merging.

PR #268, opened 2026-05-28
fix(llm): add array type guards to prevent msg.content.some crashes

apollo-mg:
## What
 Adds `Array.isArray()` type guards to `hasToolResults`, `toOpenAIUserMessage`, and `toOpenAIAssistantMessage` in the OpenAI adapter to safely handle raw string payloads.

 ## Why
 If a local model or inference server unexpectedly returns a raw string instead of a block array for `msg.content` during edge cases, the framework crashes with a fatal `msg.content.some is not a function` error. Adding these
 guards hardens the adapter against malformed histories and prevents orchestrator crashes.

 ## Checklist
 - [x] `npm run lint` passes
 - [x] `npm test` passes

### DIFF
--- a/src/llm/openai-common.ts
+++ b/src/llm/openai-common.ts
@@ -78,6 +78,8 @@ export function getOpenAIReasoningText(source: unknown): string {
  * blocks, which must be serialised as separate OpenAI `tool`-role messages.
  */
 function hasToolResults(msg: LLMMessage): boolean {
+  if (typeof msg.content === 'string') return false
+  if (!Array.isArray(msg.content)) return false
   return msg.content.some((b) => b.type === 'tool_result')
 }
 
@@ -157,6 +159,13 @@ export function toOpenAIMessages(
  * Image blocks are converted to the OpenAI image_url content part format.
  */
 function toOpenAIUserMessage(msg: LLMMessage): ChatCompletionUserMessageParam {
+  if (typeof msg.content === 'string') {
+    return { role: 'user', content: msg.content }
+  }
+  if (!Array.isArray(msg.content)) {
+    return { role: 'user', content: String(msg.content) }
+  }
+
   if (msg.content.length === 1 && msg.content[0]?.type === 'text') {
     return { role: 'user', content: msg.content[0].text }
   }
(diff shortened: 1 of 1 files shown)


JackChen-me:
The crash is real but it's a boundary problem, not an OpenAI-helper one: nothing in the framework produces a string `message.content`, so something upstream is breaking the `ContentBlock[]` contract, and the other adapters read it the same way. Send the stack trace where it throws (or point me at where you feed messages in) and I'll add a fail-fast at the entry so a bad `content` fails loudly instead of crashing deep in an adapter.

apollo-mg:
> The crash is real but it's a boundary problem, not an OpenAI-helper one: nothing in the framework produces a string `message.content`, so something upstream is breaking the `ContentBlock[]` contract, and the other adapters read it the same way. Send the stack trace where it throws (or point me at where you feed messages in) and I'll add a fail-fast at the entry so a bad `content` fails loudly instead of crashing deep in an adapter.

I had my agent dig into this a bit..

Gemini Pro 3.1: As for PR #268, Jack is 100% right. The crash isn't actually a bug in the OpenAI adapter; it's a boundary contract violation. The framework adapters expect the message.content array to be rigorously typed as a ContentBlock[] (e.g.,[{ type: 'text', text: 'Hello' }]), but something upstream is illegally passing a raw string instead. 

  I actually triggered this exact crash myself when building our consensus stress test! If you want to reply to his comment, you can give him this exact stack trace:

   1 TypeError: m.content.some is not a function
   2     at Array.some (<anonymous>)
   3     at toOpenAIMessages (src/llm/openai-common.ts:118:15)
   4     at buildOpenAIMessageList (src/llm/openai-common.ts:436:18)
   5     at OpenAIAdapter.chat (src/llm/openai.ts:142:28)
  
You can let him know that it happens when a developer (or a custom script) bypasses the Agent class and calls adapter.chat() directly while passing a raw string instead of the strict ContentBlock[] format. A fail-fast check at the entry point of adapter.chat() to reject strings would be perfect."

JackChen-me:
Merged in #288: shared `assertValidMessages` at every adapter's `chat`/`stream` entry (all providers), throwing a clear `InvalidMessageError` instead of the deep `.some` crash. Your repro is in as the regression test. Pull latest and run it against your consensus stress test, and reopen if anything slips through. Closing this in favour of #288.

PR #269, opened 2026-05-28
feat(llm): add robust regex fallback for malformed single-string JSON tools

apollo-mg:
## What
Implements a robust Regex extraction fallback in `fromOpenAICompletion` for single-string tools (like `bash` or `run_python_script`) when `JSON.parse` fails due to unescaped or triple quotes.

## Why
Highly quantized local models (like 3.5-bit or 4-bit MoEs) occasionally hallucinate Python-style triple quotes (`"""`) or unescaped quotes inside JSON tool arguments. Previously, `JSON.parse` would fail silently and       return an empty `{}` object to the model, causing it to fall into an infinite "2-Bit Drunk" hallucination loop. This fallback safely extracts the string, significantly increasing reliability and tool-calling completion rates for offline models.

## Checklist
[x] `npm run lint` passes
[x] `npm test` passes
[ ] Added/updated tests for changed behavior
[x] No new runtime dependencies (or justified in the PR description)


CodingBangboo (commented):
Thanks for your contribution @apollo-mg 

The file you changed seems to be the LLM layer and seems tool-agnostic. Would it be better to move this valid fix to the `tool` directory as a new tool? (Maybe json repair?) Plus, some tests to cover triple-quote/single and a test in which the fall-back is not fired would be great too.

Might be a bit picky but `run_python_script` is not a built-in tool.

apollo-mg:
> Thanks for your contribution @apollo-mg
> 
> The file you changed seems to be the LLM layer and seems tool-agnostic. Would it be better to move this valid fix to the `tool` directory as a new tool? (Maybe json repair?) Plus, some tests to cover triple-quote/single and a test in which the fall-back is not fired would be great too.
> 
> Might be a bit picky but `run_python_script` is not a built-in tool.

Great call on the tests! I just pushed a commit adding tests/openai-fallback.test.ts which verifies the regex correctly extracts Python triple-quotes and safely fails/bypasses on unmatchable strings.
  
Regarding run_python_script: I definitely acknowledge it's a custom tool! However, since the orchestrator relies on fromOpenAICompletion to parse the arguments before any tools are even evaluated or executed, this repair logic has
to sit in the adapter layer. The tests confirm it has zero impact on cleanly formatted tool calls, so it should be a safe fallback for anyone running local quantized models.

JackChen-me (changes requested):
The test commit overwrote the existing `tests/openai-fallback.test.ts`, dropping the 7 tests for `fromOpenAICompletion`'s text-extraction fallback (`src/llm/openai-common.ts:358-386`, still-live code whose tests going away won't fail CI). Put the new tests in their own file and leave those in place. The 3 added cases also only exercise the `"""` branch; the `'''` and `"..."`-unescape paths (the fiddliest one) are still uncovered.

On the adapter-layer point you're right, the repair has to run before tools resolve. But that doesn't require hardcoding `run_python_script`/`bash` + their param names; deriving the single-string param generically (single-key object, string value) stays in-layer and drops the name list.

apollo-mg:
> The test commit overwrote the existing `tests/openai-fallback.test.ts`, dropping the 7 tests for `fromOpenAICompletion`'s text-extraction fallback (`src/llm/openai-common.ts:358-386`, still-live code whose tests going away won't fail CI). Put the new tests in their own file and leave those in place. The 3 added cases also only exercise the `"""` branch; the `'''` and `"..."`-unescape paths (the fiddliest one) are still uncovered.
> 
> On the adapter-layer point you're right, the repair has to run before tools resolve. But that doesn't require hardcoding `run_python_script`/`bash` + their param names; deriving the single-string param generically (single-key object, string value) stays in-layer and drops the name list.

All set! I restored the original tests, implemented the generic regex parsing without hardcoded tool names, and added a dedicated test file for the quote edge cases.

JackChen-me (approved):
Generic version's right, merging. The duplicated repair block across the two parse paths is worth folding into one helper later, just a cleanup, not blocking. Thanks for sticking with the revisions.

JackChen-me:
@apollo-mg  Hey Mark, ok if I list you as a production user in the OMA README? I'd run the exact wording by you first.

apollo-mg:
Hi Jack,

That sounds great. I'm happy to be listed as a production user in the OMA
README. Please send over the wording whenever you have it ready.

Thanks,
Mark

JackChen-me:
@apollo-mg Here's the line I'd add:

> **[Mark Galyan](https://github.com/apollo-mg)** runs OMA fully offline on local quantized models, using the Coordinator and context compaction to keep autonomous agent loops alive under tight VRAM limits. Contributor since the framework's first month, across compaction, sampling, and tool-call parsing.

Swap in a different name, link, or wording if you want. Otherwise I'll add it as is.

apollo-mg:
> @apollo-mg Here's the line I'd add:
> 
> > **[Mark Galyan](https://github.com/apollo-mg)** runs OMA fully offline on local quantized models, using the Coordinator and context compaction to keep autonomous agent loops alive under tight VRAM limits. Contributor since the framework's first month, across compaction, sampling, and tool-call parsing.
> 
> Swap in a different name, link, or wording if you want. Otherwise I'll add it as is.

Sounds great, go right ahead as-is. Will DM you back on reddit once I get caught up.

JackChen-me:
@apollo-mg  Thanks! Live now, in the Built with OMA list: https://github.com/open-multi-agent/open-multi-agent#ecosystem

PR #270, opened 2026-05-28
fix(llm): prevent AgentRunner cumulative tracking bug for chunk usage

apollo-mg:
## What
Fixes an exponential token counting bug by overriding the stream chunk's input/output token assignment instead of allowing `AgentRunner` to accumulate the totals recursively.

## Why
When using the OpenAI stream API, the final chunk returns the *total* session usage. The `AgentRunner` was compounding this recursively, leading to exponentially inflated token tracking (e.g., falsely triggering context
compaction limits prematurely). This fix ensures only the actual generated tokens are recorded for that specific chunk to keep context tracking accurate.

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/src/llm/openai.ts
+++ b/src/llm/openai.ts
@@ -242,6 +242,9 @@ export class OpenAIAdapter implements LLMAdapter {
 
         // Usage is only populated in the final chunk when stream_options.include_usage is set.
         if (chunk.usage !== null && chunk.usage !== undefined) {
+          // Fix for cumulative telemetry bug: openai returns total usage for the session,
+          // but AgentRunner adds usage recursively. We should only report the tokens 
+          // generated in THIS specific stream chunk.
           inputTokens = chunk.usage.prompt_tokens
           outputTokens = chunk.usage.completion_tokens
         }


CodingBangboo (commented):
Hi @apollo-mg thanks for contributing! Before merging I wanted to flag a couple of things: 
**The diff is comments-only.** The PR title and description say this "overrides the stream chunk's input/output token assignment," but the actual code (`inputTokens = chunk.usage.prompt_tokens` / `outputTokens = chunk.usage.completion_tokens`) is unchanged — only the explanatory comment was added. 

And for the bug itself, it looks a bit confusing because:
- `chunk.usage` in the final stream chunk is the usage for **that single request**, not a session-wide total. OpenAI's API doesn't **accumulate** across requests.
- `AgentRunner` does `totalUsage = addTokenUsage(totalUsage, response.usage)` once per turn ([`runner.ts:778`](src/agent/runner.ts#L778)) — that's summing per-request usage across turns, which is the correct way to measure `maxTokenBudget` for a run. It's not recursive double-counting.

Would be awesome if we have a repro/a failing test.

CodingBangboo (changes requested):

apollo-mg:
I think I see what happened. I'm running local models via llama-server, and it turns out my local inference server has a bug where it reports the cumulative KV cache size in every chunk instead of the turn delta. Your AgentRunner logic is correct for standard OpenAI endpoints. I'll close this PR and fix it in my local proxy layer instead! Thanks for catching that.

PR #278, opened 2026-05-31
feat(shared-memory): structured handoff format for SharedMemory  

### LINKED ISSUE #17: [Feature] Structured handoff format for SharedMemory
## Summary

Support structured/typed data in SharedMemory task results, instead of plain text only. Define explicit handoff schemas so downstream agents receive well-formatted input.

## Current State

- `SharedMemory.write()` accepts only `string` values (`src/memory/shared.ts:55`)
- Comment says: *"serialise objects before writing"* — serialization is the caller's problem
- `getSummary()` outputs markdown-formatted text, truncated to 200 chars per entry
- No schema validation, no type contracts between agents

## Community Feedback

> **u/dumpshoot** (r/ClaudeAI):
> *"The multi-model setup is underrated... Match model to task type, not to perceived quality... Make handoff formats explicit... Keep the orchestration layer dumb... The value in what OP built is in the model routing logic, not the structure."*

## Proposed Approach

- Allow `SharedMemory.write()` to accept structured objects (JSON-serializable)
- Optional schema definition per handoff key (Zod schema, matching the tool framework pattern)
- Validation on write: reject malformed data before it reaches downstream agents
- Richer `getSummary()` output that respects structured data types

## Priority

**P1** — Impacts multi-agent collaboration quality. More important as workflows get complex.

btroops:
## What

This PR upgrades `SharedMemory` to support **any JSON-serializable value** (objects, arrays, numbers, booleans) instead of only strings in `write()` and `writeExpiring()`. It also adds an **optional Zod schema** to each write call to enforce type contracts and validate data at write time.

## Why

Previously, `SharedMemory` only stored strings, forcing agents to manually serialize/deserialize structured data and leaving cross-agent handoffs untyped and unsafe. This change **fixes #17** by:
- Removing manual JSON.stringify/parse boilerplate
- Ensuring type safety via Zod validation before data is persisted
- Making cross-agent handoffs explicit, typed, and less error-prone
- Preserving full backward compatibility with existing string-based usage

## Changes
-------
- MemoryEntry.value: string → unknown (types.ts)
- MemoryStore.set/setWithExpiry value param widened accordingly (types.ts)
- InMemoryStore signatures synchronized; search() adds typeof guard for non-string values (store.ts)
- SharedMemory.write/writeExpiring accept unknown value + optional ZodSchema; schema.parse(value) rejects malformed data before it reaches the store
- getSummary() renders objects as pretty-printed JSON, primitives via String(), and truncates at 200 chars (shared.ts)
- 20 new tests covering structured values, backward compat, getSummary formatting, and Zod validation (tests/shared-memory.test.ts)
- Docs updated with typed handoff, schema validation, and summary examples

Closes #17

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/docs/shared-memory.md
+++ b/docs/shared-memory.md
@@ -57,7 +57,7 @@ await mem.write('counter', 'remaining', 7)
 await mem.write('checker', 'valid', true)
 ```
 
-The underlying `MemoryStore` always stores values as strings (via `JSON.stringify`). On read, `entry.value` is typed as `unknown` — downstream callers should narrow it with a type guard or `JSON.parse()`.
+`SharedMemory` serialises values to JSON strings before writing to the underlying `MemoryStore` and deserialises them back to their original type on read, so downstream callers receive well-typed data without manual parsing.
 
 ### Schema validation (Zod)
 
--- a/docs/shared-memory.md
+++ b/docs/shared-memory.md
@@ -25,3 +25,84 @@ const team = orchestrator.createTeam('durable-team', {
 ```
 
 When both are provided, `sharedMemoryStore` wins. SDK-only: the CLI cannot pass runtime objects.
+
+---
+
+## Structured handoff (typed values)
+
+`SharedMemory.write()` and `writeExpiring()` accept any JSON-serializable value, not just strings:
+
+```typescript
+const mem = new SharedMemory()
+
+// String (backward-compatible)
+await mem.write('researcher', 'findings', 'TypeScript 5.5 ships const type params')
+
+// Object (auto-serialized via JSON.stringify)
+await mem.write('analyst', 'summary', {
+  total: 42,
+  passed: 40,
+  failed: 2,
+  tags: ['regression', 'critical'],
+})
+
+// Array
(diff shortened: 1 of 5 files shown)


JackChen-me:
Nice work, matches the reopened #17  scope.

Before merge though: docs/shared-memory.md says values are stored as strings via JSON.stringify and narrowed with a type guard or JSON.parse on read, but write() stores them raw, so the documented JSON.parse path throws a SyntaxError. Same gap on the write side: MemoryStore.set now takes unknown with nothing serializing, so a string-assuming Redis or SQLite backend persists [object Object].

We'd serialize at the SharedMemory boundary (stringify on write, parse on read) and keep the store string-only, rather than storing raw, so durable backends keep getting strings. Parse needs a try/catch fallback for non-JSON values written via getStore(). Heads up it's more than a docs change: the MemoryStore.set signature and search() guard come back out, and getSummary reworks its type branch.

Minor: the truncation test only checks the - big: line, which is just - big: { after pretty-printing, so it passes either way. Assert on the … marker instead.

PR #279, opened 2026-05-31
fix(shared-memory): serialize values at boundary to keep store string-only

### LINKED ISSUE #17: [Feature] Structured handoff format for SharedMemory

btroops:
## What

Serialize values at the `SharedMemory` boundary to keep `MemoryStore` string-only. All writes use `JSON.stringify`, and all reads use `JSON.parse` with a safe fallback.

## Why

This fixes a critical mismatch between documentation and implementation, and ensures durable backends (Redis, SQLite) do not receive raw objects that become `[object Object]`. It maintains backward compatibility and aligns with the original design of string-only storage.

Fixes #17

## Changes

- Revert `MemoryStore.set/setWithExpiry` to accept `string` values (types.ts)
- Update `InMemoryStore` signatures; `search()` uses string casting (store.ts)
- `SharedMemory.write()` and `writeExpiring()` now `JSON.stringify` before storing
- `SharedMemory.read()` safely parses values with `try/catch JSON.parse`
- Add `parseStoredValue()` helper for backward compatibility with direct store writes
- `getSummary()` uses parsed values for correct formatting and truncation
- Truncation test now asserts on the `…` marker for valid coverage
- Update docs to reflect automatic parsing (users no longer need manual `JSON.parse`)

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] Addressed all review feedback
- [x] Backward compatible

### DIFF
--- a/docs/shared-memory.md
+++ b/docs/shared-memory.md
@@ -57,7 +57,7 @@ await mem.write('counter', 'remaining', 7)
 await mem.write('checker', 'valid', true)
 ```
 
-The underlying `MemoryStore` always stores values as strings (via `JSON.stringify`). On read, `entry.value` is typed as `unknown` — downstream callers should narrow it with a type guard or `JSON.parse()`.
+`SharedMemory` serialises values to JSON strings before writing to the underlying `MemoryStore` and deserialises them back to their original type on read, so downstream callers receive well-typed data without manual parsing.
 
 ### Schema validation (Zod)
 
--- a/docs/shared-memory.md
+++ b/docs/shared-memory.md
@@ -25,3 +25,84 @@ const team = orchestrator.createTeam('durable-team', {
 ```
 
 When both are provided, `sharedMemoryStore` wins. SDK-only: the CLI cannot pass runtime objects.
+
+---
+
+## Structured handoff (typed values)
+
+`SharedMemory.write()` and `writeExpiring()` accept any JSON-serializable value, not just strings:
+
+```typescript
+const mem = new SharedMemory()
+
+// String (backward-compatible)
+await mem.write('researcher', 'findings', 'TypeScript 5.5 ships const type params')
+
+// Object (auto-serialized via JSON.stringify)
+await mem.write('analyst', 'summary', {
+  total: 42,
+  passed: 40,
+  failed: 2,
+  tags: ['regression', 'critical'],
+})
+
+// Array
+await mem.write('tester', 'reports', [
+  { id: 1, status: 'pass' },
+  { id: 2, status: 'fail', reason: 'timeout' },
+])
+
+// Number / boolean
+await mem.write('counter', 'remaining', 7)
+await mem.write('checker', 'valid', true)
+```
+
+The underlying `MemoryStore` always stores values as strings (via `JSON.stringify`). On read, `entry.value` is typed as `unknown` — downstream callers should narrow it with a type guard or `JSON.parse()`.
+
+### Schema validation (Zod)
+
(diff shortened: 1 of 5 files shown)


CodingBangboo (commented):
Thanks for the contribution! @btroops
```
if (schema) {
  schema.parse(value)
}
const serialised = JSON.stringify(value) ?? 'null'
```
In short, if I'm understanding it correctly,`schema.parse(value)` returns the parsed value but the code ignores the return and `stringifies` the original value. 
So when a schema changes the data during parsing, that change is lost. (But if that parse is an inplace change then ignore my comment)

Two common cases:

1. Coercion — z.object({ n: z.coerce.number() })

await mem.write('a', 'k', { n: '42' }, undefined, schema)
// schema.parse turns n into the number 42, but that result is dropped.
// What's actually stored: { n: '42' }  ← still a string

2. Defaults — z.object({ active: z.boolean().default(true) })
await mem.write('a', 'k', {}, undefined, schema)
// schema.parse would produce { active: true }, but that's dropped.
// What's actually stored: {}  ← default never applied

The fix is to store what parse returns, not the original:
```
const validated = schema ? schema.parse(value) : value
const serialised = JSON.stringify(validated) ?? 'null'
```

And it slips through CI because the PR's schema tests only use plain validation (right type passes, wrong type throws) — none use `.coerce`, `.default()`, or `.transform()`, so the discarded-result bug is never exercised.

JackChen-me:
Three things before merge.

Biggest is where the typed value lives: MemoryEntry.value is unknown, but the store only ever holds strings (hence the as string cast in search()). That typed semantic belongs on the read surface, not the store layer: keep MemoryEntry.value: string (e.g. MemoryEntry<T = string>) and return MemoryEntry<unknown> from read()/listAll()/listByAgent(). That leaves store.get() consumers and custom MemoryStore implementers untouched, drops the cast, and scopes the break to typed-read consumers (worth a changelog note, since it's not fully backward compatible).

Same surface, runtime gap: read() deserializes but listAll()/listByAgent() don't, so a string write comes back as '"foo"' instead of 'foo' there, a silent regression. They need parseStoredValue too.

And @CodingBangboo's point above is right. Store what parse returns.

btroops:
Thanks a lot for the detailed reviews from both of you! I'll fix all three points in the next commit.

JackChen-me:
@btroops  Still planning to push the follow-up on the three fixes, or want me to take it from here? Happy to land it either way.

JackChen-me:
Superseded by [#284](https://github.com/open-multi-agent/open-multi-agent/pull/284). Direction was right, same shape that shipped. Thanks for the time on it @btroops

Issue #275, opened 2026-05-29
[Proposal] Built-in adversarial verification / consensus primitive (propose, refute, converge)


JackChen-me:
## What

A built-in primitive for "propose N independent answers, have other agents refute them, iterate until they converge," instead of every team hand-wiring reviewer/voter agents into a task graph.

## Why

OMA can do this manually today (wire reviewer agents, aggregate structured outputs in user code), but there's no built-in pattern. Code check: no consensus/debate/verify-until-converge helper; `Scheduler`'s capability-match is task-to-agent assignment, not verification. Anthropic's Dynamic Workflows ships this as a headline feature, so it's becoming an expected capability rather than a niche. In practice the shape holds up: one set of agents proposes, a second adversarially refutes each, then synthesize the survivors, which lands materially more grounded than a single pass.

## Rough shape (for discussion)

Something like `consensus(attempts, { judges, mode: 'refute' | 'lens' })` composing `AgentPool.runParallel` + structured output, returning a verdict plus the dissent. Two verifier styles worth supporting: N identical skeptics (majority refutes, kill) and diverse lenses (each verifier a different angle), caller's choice. Should compose with `runTeam` / `runTasks` so a task can be marked "verify before accepting." Opt-in per task, no hard-coded vote threshold.

## Status

Sourced from the Dynamic Workflows competitive analysis plus one internal use, not committed. No external user has asked yet, so it needs demand signal before scheduling. If you'd use a built-in verify/consensus step, say what shape and where in the pipeline.

CodingBangboo:
Direction looks great to me, but just a few things before we get started:
1. **One-shot tasks shouldn't pay for this.** Lots of tasks have a single right answer and don't need verification. The primitive should be something a task opts *into*, not a new default mode — keep `runTask` clean.
2. **"Iterate until converge" needs hard bounds.** Without a cap, this is a recipe for surprise bills. I'd suggest:
   - `maxRounds` (default 2: propose + one refute) — no unbounded loops.
   - Early exit on quorum (K of N skeptics accept) so we don't run every verifier.
   - Token accounting must roll up into the parent's `totalUsage` / `maxTokenBudget`, the same way `delegate_to_agent` does today. Otherwise consensus silently bypasses the parent budget.
 
### Proposed shape

A composable function, not a new top-level mode:

```ts
runConsensus(team, prompt, {
  proposer,                  // single agent or N-best
  judges,                    // verifier roster
[...]

PR #280, opened 2026-06-01
feat(orchestrator): add consensus / adversarial verification primitive

CodingBangboo:
## What

Adds a built-in **propose → refute → converge** primitive so users get grounded, scrutinised answers without hand-wiring reviewer agents into a task graph.
- **`OpenMultiAgent.runConsensus(team, prompt, options)`** — a proposer emits an answer; a roster of judges tries to refute it over up to `maxRounds`, exiting early once a `quorum` accepts.
- **Per-task `verify` hook** — any `runTasks` task can opt into the same loop on its own result; a revised answer supersedes the task output for downstream consumers.

## Merge Gate Mentioned in the Issue

Proposer/judge/revision usage all accumulate into the run's cumulative usage and are checked against `maxTokenBudget`: crossing it stops further judge calls, mirroring `delegate_to_agent`

## Why

Follows #275 

Just to reiterate the API shape and the opt-in verification of the task mentioned in #275's discussion:
```ts
const result = await orchestrator.runConsensus(team, 'Is this proof correct?', {
  proposer: { name: 'solver', model: 'claude-opus-4-6' },   // or an array (N-best)
  judges: [judgeA, judgeB],
  mode: 'refute',       // 'refute' (identical skeptics) | 'lens' (distinct angles)
  quorum: 2,            // default: ceil(judges.length / 2)
  maxRounds: 2,         // default: 2
  onDissent: 'revise',  // 'revise' | 'reject' | 'keep'
})
```
```
// Opt-in verification of a pipeline task's own result:
await orchestrator.runTasks(team, [
  { title: 'derive-bound', description: '…', assignee: 'mathematician',
    verify: { judges: [judgeA, judgeB], maxRounds: 2 } },
])
```

## Additional design notes
1. Judges run on the existing `AgentPool` via `runEphemeral`, and verdicts reuse the structured-output path (`extractJSON` + `validateOutput`) to parse the `{ accept, critique }` contract.
2. Sequential judges, not parallel
3. The per-task `verify` hook is awaited inside the task's dispatch promise, before the wave's `Promise.all(dispatchPromises)` resolves. A revised answer is written back to the queue (`queue.update(taskId, { result })`) and shared memory before any dependent task is dispatched, so downstream consumers always read the post-verification result, never the pre-revision one.
4. `quorum` defaults to `ceil(judges.length / 2)` and is fully configurable;`maxRounds` defaults to `2`. Nothing about the vote count is baked in.
5. Verification modes
- **`refute`** — every judge receives the same skeptic framing.
- **`lens`** — each judge is assigned a distinct review angle (correctness, completeness, edge cases, clarity, assumptions, evidence) round-robin by index, with a built-in instruction to accept rather than invent objections when the
  angle is irrelevant to the question. 

`judgePrompt` overrides the framing entirely — a `string` for all judges, or a `(judge) => string` function for per-judge prompts. Every judge prompt includes the original question alongside the proposed answer.Users can input their own instructions like this:
```
// Define your judges
const judges = [
  { name: 'security',    model: 'claude-opus-4-6' },
  { name: 'performance', model: 'claude-opus-4-6' },
  { name: 'api-design',  model: 'claude-sonnet-4-6' },
]

// Your own angles, keyed by judge name
const angles: Record<string, string> = {
  security:    'Focus only on security: injection, authz, secret handling, unsafe deserialization.',
  performance: 'Focus only on performance: N+1 queries, allocations in hot paths, blocking I/O.',
  'api-design':'Focus only on API ergonomics: naming, backward compatibility, error contracts.',
}

const result = await orchestrator.runConsensus(team, 'Review this PR diff: …', {
  proposer: { name: 'author', model: 'claude-opus-4-6' },
  judges,
  judgePrompt: (judge) => angles[judge] ?? 'Review the answer for correctness.',
})
```
## Some good follow-up works
1. add `examples/patterns/consensus.ts`
2. `runTeam` integration. The `verify` hook is wired into explicit `runTasks` tasks. Coordinator-generated tasks in `runTeam` can't yet opt in (the coordinator emits task JSON without a verify field)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,8 @@ Local models via Ollama need no API key, see [`providers/ollama`](examples/provi
 | Auto-orchestrated team | `runTeam()` | Give a goal, let the coordinator plan and execute | [`basics/team-collaboration`](examples/basics/team-collaboration.ts) |
 | Explicit pipeline | `runTasks()` | You define the task graph and assignments | [`basics/task-pipeline`](examples/basics/task-pipeline.ts) |
 
+For answers that need scrutiny, `runConsensus()` runs a proposer→judge verification loop (with an opt-in per-task `verify` hook). See [Consensus](./docs/consensus.md).
+
 Preview the coordinator's task DAG without executing agents:
 
 ```ts
@@ -374,6 +376,7 @@ Before going live, wire up the controls that protect token spend, recover from f
 - [Shared memory](https://github.com/open-multi-agent/open-multi-agent/blob/main/docs/shared-memory.md) — the default store and custom `MemoryStore` backends.
 - [Context management](https://github.com/open-multi-agent/open-multi-agent/blob/main/docs/context-management.md) — sliding window, summarization, compaction, and custom compressors.
 - [CLI](https://github.com/open-multi-agent/open-multi-agent/blob/main/docs/cli.md) — the JSON-first `oma` binary for shell and CI.
+- [Consensus](./docs/consensus.md) — the `runConsensus` proposer→judge primitive, the per-task `verify` hook, and the budget invariant.
 
 ## Contributing
 
--- a/docs/consensus.md
+++ b/docs/consensus.md
@@ -1,0 +1,63 @@
+# Consensus
+
+`runConsensus` adds a proposer→judge verification loop on top of a single prompt: a **proposer** agent emits an answer, then a roster of **judge** agents try to refute it over up to `maxRounds`. The loop exits early once a `quorum` of judges accept.
+
+```ts
+const result = await orchestrator.runConsensus(team, 'Is this proof correct?', {
+  proposer: { name: 'solver', model: 'claude-opus-4-6' },
+  judges: [
+    { name: 'judge-a', model: 'claude-opus-4-6' },
+    { name: 'judge-b', model: 'claude-sonnet-4-6' },
+  ],
+  mode: 'refute',       // identical skeptic framing for every judge
+  quorum: 2,            // default: ceil(judges.length / 2)
+  maxRounds: 2,         // default: 2
+  onDissent: 'revise',  // default: feed dissent back to the proposer
+})
+
+result.answer      // the (possibly revised) answer
+result.verdict     // 'accepted' | 'rejected'
+result.dissent     // critiques recorded across all rounds
+result.rounds      // judging rounds executed
+result.tokenUsage  // proposer + judges + revisions
+```
+
+## Options
+
+| Option | Default | Meaning |
+|--------|---------|---------|
+| `proposer` | — | One agent, or an array (N-best — all candidates are shown to the judges). |
+| `judges` | — | Verifier roster. Judges run **sequentially** so quorum and budget can stop the rest. |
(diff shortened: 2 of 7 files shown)


JackChen-me (commented):
`runTaskVerify` (`src/orchestrator/orchestrator.ts:1146`) never reads `consensus.verdict`. The task is already completed when the hook runs, and the swap at `:1195` fires on any changed answer regardless of verdict, so with the default `onDissent: 'revise'` a revision the judges rejected still overwrites the task output and shared memory (`:1196-1197`) downstream. It can't actually reject anything yet. Fix: swap only when `verdict === 'accepted'`, and surface the verdict as a task-level outcome.

The reviser also never sees the answer it's revising: `buildRevisePrompt` (`:984`, called at `:1127`) gets the prompt + critiques but not the in-scope `answer`, and the proposer runs through `runEphemeral` with no memory, so it re-derives from scratch instead of patching. Pass `answer` in. Neither surfaces in the tests because the mock proposer ignores its prompt.

Minor: `quorum` isn't capped at `judges.length` (`:1821`, `:1166`), so an over-set quorum just burns rounds before rejecting. Rest looks right; CI's green. The two correctness items are what I'd fix before merge.

CodingBangboo:
@JackChen-me Thanks for the comments. In the new commit I've fixed three things.
1. Verdict-gated swap + surface verdict as task outcome
2. Reviser sees the answer it's revising
3. Quorum capped at judges.length

let me know if we need to fix more

apollo-mg:
This is an excellent idea.

JackChen-me:
Two more, both correctness, both silent (the reported result doesn't match what actually ran). The bigger one breaks the payoff of `verify`: when judges accept a revision, the improved answer reaches downstream tasks but never the caller. `runTaskVerify` runs after the task's finalized (`queue.complete()` + the `task_complete`/`agent_complete` events at `:767`+, verify at `:787`), and the accepted swap only touches the queue and shared memory (`:1211`) while `ctx.agentResults` keeps the old output (`:1192` bumps `tokenUsage` only) and `taskRecords` has no result field (`:1798`). So the returned `TeamRunResult`, the progress events, and the approval gate all show the pre-verify answer, and on a terminal verified task the improvement vanishes from the caller entirely. Running verify before `queue.complete()`/the events fixes it, so the task finalizes with the verified result (accepted → revised, rejected → original as today).

Second, `accepted` can't be trusted when the proposer produces nothing: `runConsensus()` can return `{ verdict: 'accepted', answer: '' }`. Only `r.success && r.output` gets pushed to `candidates` (`:1862`), with no empty-`candidates` guard before the judge loop runs on the empty `initialAnswer` (`:1883`), so a failed proposer is swallowed and can come back accepted, indistinguishable from a real rejection. Bail when `candidates` is empty, same shape as the budget-exceeded early return right above.

CodingBangboo:
Thanks @JackChen-me 
I've looked at your comments and they look very reasonable. Fixes are been commited.

JackChen-me:
Merged main in to clear the conflicts, just an import line and a README block. Lint and tests green locally. Merging once CI confirms.

PR #281, opened 2026-06-01
feat(llm): add Tencent Hunyuan provider adapter

KaitlynFeng:
## What

Adds a built-in `hunyuan` provider shortcut for the OpenAI-compatible Tencent Hunyuan API, following the existing `doubao`/`qiniu` thin-wrapper pattern. Includes adapter, factory registration, CLI reference, unit tests, an example, and docs.

## Why

Hunyuan is a widely used Chinese LLM family that wasn't yet available as a built-in shortcut. It speaks the OpenAI Chat Completions wire format, so it fits the existing adapter model cleanly.

One nuance worth flagging for reviewers: Tencent exposes Hunyuan through **two independent OpenAI-compatible surfaces with separate API-key namespaces**:

- **Tencent MaaS / TokenHub** (`https://tokenhub.tencentmaas.com/v1`, `sk-...` keys, models like `hy3-preview`) — Tencent's **current** platform.
- **Legacy Tencent Cloud** (`https://api.hunyuan.cloud.tencent.com/v1`, console keys, models like `hunyuan-turbos-latest`) — Tencent has officially announced this platform is being **retired** (sales stop 2026-06-30, full shutdown 2026-09-30).

Because the legacy endpoint is on its way out, the adapter **defaults to the TokenHub endpoint** and lets users target the legacy endpoint via `HUNYUAN_BASE_URL` until it is shut down. This mirrors the existing `MIMO_BASE_URL` / `MINIMAX_BASE_URL` convention. Happy to flip the default if maintainers prefer.

### Verification

Verified **chat + function calling end-to-end through the framework adapter interface** (normalized `tool_use` blocks, not just raw HTTP) against both surfaces:

- `hy3-preview` on TokenHub
- `hunyuan-turbos-latest` on the legacy Tencent Cloud endpoint

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes (964 tests, incl. new `tests/hunyuan-adapter.test.ts`)

Made with [Cursor](https://cursor.com)


JackChen-me:
Clean, follows the doubao/qiniu pattern. One real question before merge: verification covers hy3-preview and turbos, but the docs list tool calling on hunyuan-t1 too, and t1 is a reasoning model. That's where deepseek/mimo needed echoesReasoning: 'tool-use-only' (reasoning_content must be echoed back on every tool-calling turn or the next 400s); HunyuanAdapter inherits the default 'never', so if t1 follows that contract multi-turn tool calls on it break. Did you test t1? If not, simplest to narrow the docs to what you verified.

TokenHub default is the right call, no need to flip it.

KaitlynFeng:
Thanks Jack. Tested t1 — its multi-turn tool calling is flaky for a different reason (it ignores `role:"tool"` content on the legacy endpoint, unrelated to reasoning replay), so I just narrowed the docs to what's verified (hy3-preview / turbos / functioncall) as you suggested.

But your instinct was right for the **default** model: hy3-preview is itself a reasoning model, and Tencent's TokenHub spec requires `reasoning_content` to be backfilled on every tool turn in interleaved-thinking mode ([Interleaved Thinking](https://cloud.tencent.com/document/product/1823/130930), [call guide](https://cloud.tencent.com/document/product/1823/132252)). So I set `HunyuanAdapter.capabilities = { echoesReasoning: 'tool-use-only' }` (same as DeepSeek/MiMo) and verified end-to-end on TokenHub — hy3-preview at `reasoning_effort: high` runs a 3-step tool loop with reasoning echoed each turn, no regression. turbos/functioncall emit no `reasoning_content`, so it's a no-op for them. Added unit coverage; TokenHub stays default.

PR #283, opened 2026-06-02
feat: add LiteLLM as AI gateway provider

RheagalFire:
## What

Adds LiteLLM as a provider (`provider: 'litellm'`), giving users access to 100+ LLM providers through a [LiteLLM proxy](https://github.com/BerriAI/litellm) server. Thin `OpenAIAdapter` wrapper following the same pattern as DeepSeek, Grok, and MiniMax adapters.

## Why

Users who run a LiteLLM proxy can route to Anthropic, OpenAI, Gemini, Azure, Bedrock, Ollama, and 100+ other providers from a single OpenAI-compatible endpoint. This adapter lets them use `provider: 'litellm'` instead of manually configuring `baseURL` on the OpenAI adapter.

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes (978 passed, 0 failures)
- [x] Added/updated tests for changed behavior (6 tests in `tests/litellm-adapter.test.ts`)
- [x] No new runtime dependencies (reuses `openai` SDK already in deps; lazy import)


RheagalFire:
cc @JackChen-me

JackChen-me:
Works today without an adapter: the LiteLLM proxy is OpenAI-compatible, so `provider: 'openai'` + `baseURL` reaches it, same as OpenRouter and the other OpenAI-compatible endpoints in `docs/providers.md`. It's self-hosted with no canonical endpoint, so a built-in would add a provider name but no usable default (just your own `localhost:4000`). Better as a row in that table. Want to send that instead?

JackChen-me:
@RheagalFire   Offer's still open: convert this to a `docs/providers.md` row like the OpenRouter one and I'll merge it. Otherwise I'll close in a few days.

RheagalFire:
> @RheagalFire Offer's still open: convert this to a `docs/providers.md` row like the OpenRouter one and I'll merge it. Otherwise I'll close in a few days.

Thanks i missed earlier notification, I'll make this change.

RheagalFire:
Updated - converted to a docs/providers.md row.

chengduman:
Nice addition — LiteLLM makes sense as a provider adapter. A few production considerations from running LiteLLM as a multi-agent gateway:

## Cost routing

LiteLLM's `routing_strategy="latency-based-routing"` is useful but incomplete for multi-agent systems. Consider documenting a **cost-aware routing** pattern where the OMA agent picks the cheapest provider that meets the task's complexity threshold:

```yaml
# Example cost-routing config
litellm_routing:
  default: deepseek-chat
  fallback_order: [claude-haiku, gpt-4o-mini, gpt-4o]
  max_cost_per_task: 0.05
  retry_on_error: true
  retry_with_higher_model: true
```

## Dual adapter pattern

Users running LiteLLM often also run direct provider adapters (DeepSeek, Groq, Ollama). There's a subtle interaction: if both `litellm` and `deepseek` are configured, which wins? A "prefer direct, fallback to LiteLLM" pattern would be worth documenting in the cookbook.

## Rate limits

LiteLLM's default exponential backoff doesn't account for OMA's retry logic. Worth noting that retries should be configured on ONE side — otherwise you get compounding delays.

Would it make sense to add a `routing` config block to the provider definition so users can express these preferences declaratively?

PR #286, opened 2026-06-03
feat: add deterministic model routing policy

### LINKED ISSUE #271: [Proposal] Cost/capability-aware model routing (cheap models for fan-out, premium for synthesis)
## What

Optional model routing so a task can run on a cheap model for high-volume or low-stakes work and a premium one for synthesis or critical-path steps, instead of every agent being pinned to a single `model`.

## Why

Multi-provider is the framework's structural strength, but today it's only "you can mix" via per-`AgentConfig` `model`; nothing picks cheap vs. premium automatically. The `Scheduler`'s `capability-match` matches tasks to agents by keyword, not models by cost. Closest user signal is u/dumpshoot in #17: "match model to task type, not to perceived quality." And single-vendor orchestration is structurally token-heavy, so cost routing across providers is something only a multi-provider runtime can offer.

## Rough shape

A `routing` policy (a cheap/premium pair, escalate on low confidence, or route by task tag or leaf-ness) that composes with the existing per-agent `provider`/`model` and the AI SDK adapter. Declarative heuristics first, no learned router. It only ever picks among models the user explicitly allows, never auto-selects unconfigured ones.

## Status

From competitive analysis, not committed. Want real demand before scheduling: if you'd use this, say how you'd route. The dumpshoot signal is supportive but thin.

cat0825:
## Summary
- add an opt-in deterministic model routing policy for team orchestration
- support predicate-free routing by phase, agent, task role, priority, leaf status, and dependency status
- apply routes without mutating team/agent configuration
- route worker, delegated, short-circuit, coordinator, and synthesis calls through the same policy shape

## Tests
- `node ./node_modules/vitest/dist/cli.js --run tests/orchestrator.test.ts tests/agent-pool.test.ts tests/task-utils.test.ts`
- `npm run lint`

Fixes #271


JackChen-me:
CI's red: the per-task `buildAgent()` + `runEphemeral` bypasses the pooled agent and breaks the 9 `onApproval` tests. Build the routed agent only when a rule matches, otherwise keep `pool.run(assignee)` so the default path is unchanged.

PR #292, opened 2026-06-07
feat(minimax): upgrade default model to MiniMax-M3

octo-patch:
## Summary

Promote `MiniMax-M3` to the default model used by the `minimax` provider. `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` remain available for callers that pin a model explicitly.

## Changes

- `src/cli/oma.ts` — default model hint for `minimax` switched from `MiniMax-M2.7` to `MiniMax-M3`.
- `src/llm/minimax.ts` — adapter docstring updated to reference M3 (no runtime behavior change; adapter remains model-agnostic).
- `examples/providers/minimax.ts` — all three example agents and `defaultModel` switched to `MiniMax-M3`.
- `docs/providers.md`, `docs/providers/minimax.md`, `examples/README.md`, `README.md`, `README_zh.md` — model recommendation updated.

The adapter itself does not enumerate model IDs, so M2.7 / M2.7-highspeed continue to work via the existing `provider: 'minimax'` + `model: '<id>'` config — no model list to maintain.

## Why

`MiniMax-M3` is MiniMax's latest flagship model: 512K context window, up to 128K output tokens, and image input support on the OpenAI-compatible endpoint. M2.7 (192K context) is the previous generation.

## Test plan

- [x] `npm run lint` passes
- [x] `npm test -- minimax` passes (6 tests in `tests/minimax-adapter.test.ts`)
- [x] Existing tests are model-agnostic — no test updates needed since the adapter only depends on `apiKey` / `baseURL`


JackChen-me:
Merged. Edited one line in the M3 doc before merging: M3's context is 1M (MiniMax lists 512K as the guaranteed floor, not the cap), and I dropped the "128K output tokens" number since it's not in MiniMax's published model docs.

Issue #20, opened 2026-04-02
[Feature] State persistence and checkpoint/resume for long-running workflows


JackChen-me:
## Summary

Add state persistence and checkpoint/resume capability so that long-running multi-agent workflows can survive process restarts and scale beyond in-memory limits.

## Current State

- All state is in-memory: `InMemoryStore` (comment: *"never persisted to disk"* — `src/memory/store.ts:4`), `TaskQueue`, agent conversation history
- Process crash = all progress lost, no recovery possible
- Coordinator is stateless one-shot (generates task JSON, then done)
- `AgentRunner` supports `AbortSignal` for cancellation, but no resume after abort
- No checkpoint mechanism at any level (task, agent, or workflow)

## Community Feedback

> **u/mrtrly** (r/LocalLLaMA):
> *"Extracted code gets messy fast when the original design had context you don't see in the source. The orchestration layer is clean until you hit the edge cases... you'll spend way more time on coordinator state management than the code suggests once you're actually shipping multi-agent work at scale."*

> **u/ExplorerPrudent4256** (r/LocalLLaMA):
> *"You'd need timeout recovery, state persistence across agents, and a strategy for partial failures in the task graph. Honestly, the coordination overhead kills you. More agents = exponentially more state to track. That's why most local implementations just end up being single-agent with better tooling."*

## Proposed Approach

- Pluggable `StateStore` interface (in-memory default, file/SQLite/Redis optional)
- Checkpoint after each task completion: save task results, SharedMemory state, queue state
- Resume from checkpoint: reconstruct workflow state, skip completed tasks, restart failed/pending
- Workflow-level `save()` / `restore()` API
- Consider event sourcing: persist each state transition, reconstruct by replaying

## Priority

**P3** — Important for production scale, but current users are primarily local dev / small workflows. Build this when real demand appears.

JackChen-me:
**Source:** Reddit community feedback
- u/mrtrly
- u/ExplorerPrudent4256

JackChen-me:
Reopening. Anthropic's Dynamic Workflows ships durable resumable runs as a headline feature, which moves whole-run checkpoint/resume from a scale-time nicety to a category baseline. We triaged this P3 ("build when real demand appears"); that gate's met now: the original feedback here, plus it being table-stakes at the category leader, plus a fresh check confirming run-level resume is still absent (only task-level retry exists, `InMemoryStore` is never persisted).

[...]

PR #294, opened 2026-06-08
feat: checkpoint/resume first slice (per-run checkpoint over MemoryStore)

mvanhorn:
## Summary
Adds durable run-level checkpoint/resume so a long-running workflow can be saved and continued, built entirely on the existing `MemoryStore` API.

## Why this matters
Requested in #20 (reopened, bumped to P1): only task-level retry existed, with no way to persist and resume a whole run. This adds a first slice: a `Checkpoint` manager that snapshots run state to the memory store under a reserved key namespace, and orchestrator/queue integration to save on progress and restore on start. Because it depends only on the public `MemoryStore` interface, in-memory, Redis, SQLite, and custom backends all work without extra hooks.

## Changes
- New `memory/checkpoint.ts`: `Checkpoint` class, namespaced keys, snapshot save/load.
- `types.ts`: `CheckpointOptions`, `CheckpointSnapshot`, `RestoreOptions`.
- Orchestrator + task queue: save a checkpoint as tasks complete; restore from a snapshot on resume.

## Scope
This is the first slice: per-run snapshot/restore over `MemoryStore`. Backend-specific durability guarantees and finer-grained mid-task checkpointing are intentionally left for follow-ups so the core API can be reviewed first.

Refs #20

AI was used for assistance.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -121,6 +121,13 @@ export { TokenBudgetExceededError, InvalidMessageError } from './errors.js'
 
 export { InMemoryStore } from './memory/store.js'
 export { SharedMemory } from './memory/shared.js'
+export {
+  Checkpoint,
+  CHECKPOINT_KEY_PREFIX,
+  DEFAULT_CHECKPOINT_KEY,
+  checkpointKey,
+  isCheckpointKey,
+} from './memory/checkpoint.js'
 
 // ---------------------------------------------------------------------------
 // Types — all public interfaces re-exported for consumer type-checking
@@ -168,6 +175,8 @@ export type {
   TeamRunResult,
   RunTeamOptions,
   RunTasksOptions,
+  RunTaskSpec,
+  RestoreOptions,
   ModelRouteConfig,
   ModelRoutingMatch,
   ModelRoutingRule,
@@ -192,6 +201,11 @@ export type {
   OrchestratorConfig,
   OrchestratorEvent,
   CoordinatorConfig,
+  CheckpointOptions,
+  CheckpointSnapshot,
+  CompletedTaskCheckpoint,
+  TaskQueueSnapshot,
+  TaskSnapshot,
 
   // Trace
   TraceEventType,
@@ -208,6 +222,8 @@ export type {
   // Memory
   MemoryEntry,
   MemoryStore,
+  MemoryEntrySnapshot,
+  SharedMemorySnapshot,
   SharedMemoryEntry,
   SharedMemoryValue,
   SharedMemoryWriteOptions,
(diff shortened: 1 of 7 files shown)


JackChen-me:
This is clean, suite passes. Couple things, one worth settling first:

The big one: a failed checkpoint write takes the whole run down. `saveRunCheckpoint` is awaited inside the dispatch map, so a rejecting `store.set` rejects the `Promise.all` and bubbles out mid-batch. Inverts the point of checkpointing the moment a real store (Redis/SQLite) hiccups. Best-effort + surface via `onProgress`, or fail-fast documented explicitly? Either way a failing-store test would pin it, since `InMemoryStore` never rejects.

Smaller, all defer-able: `runTeam` restore skips the final synthesis (`executeExplicitTaskQueue` path), so a restored team run returns raw task outputs, not the synthesized answer; worth a line in the `restore` JSDoc. No `docs/` for the new surface yet. And the per-task full shared-memory snapshot is ~O(N²) write volume, fine for a first slice but a follow-up.

CodingBangboo (commented):
Thanks for the contribution! @mvanhorn 

**`fallbackCheckpointStore` is instance-shared — concurrent runs stomp each other**

`orchestrator.ts:1421` — `private readonly fallbackCheckpointStore = new InMemoryStore()` is created once per `OpenMultiAgent` instance. When `checkpoint: true` is passed without a `store` or `runId`, `createActiveCheckpoint` falls back to this store. Two concurrent runs on the same instance — which `runTeam`/`runTasks` supports — will both write to `DEFAULT_CHECKPOINT_KEY = '__oma_checkpoint__/latest'`, silently overwriting each other's checkpoints.

**Fix:** create a per-call fallback store, or require `runId` when `checkpoint: true` and no `store` is provided.

AI assisted in this PR review.

JackChen-me:
@mvanhorn the call on both, so they're not blocking you:

Write failure: go best-effort. Catch `save` and surface via `onProgress` instead of letting it reject out of `executeQueue`. A checkpoint write shouldn't be able to kill the run it protects.

Fallback store: require a `runId` (or explicit `store`) when `checkpoint` is on. Per-call dodges the stomp but can't resume across a restart, so it misses the point.

Got time to pick these up?

JackChen-me:
Applied the two we landed on directly: best-effort checkpoint saves (a failing store surfaces via `onProgress`, plus a test) and a required `runId`/`store` before the in-memory fallback. Synced and merging. Docs and the O(N²) snapshot volume can come as follow-ups.

mvanhorn:
Thanks @JackChen-me for merging the checkpoint work. Durable run-level checkpoint and resume is a big step for long workflows.

JackChen-me:
@mvanhorn  Glad it landed. You've gone from the April examples to the checkpoint internals, so what are you running that actually needs durable resume?

mvanhorn:
Thank you @JackChen-me. Durable checkpoint/resume built on the existing MemoryStore is a strong first slice.

PR #297, opened 2026-06-16
feat(examples): add consensus pattern example

nuthalapativarun:
## What

Adds `examples/patterns/consensus.ts` — a runnable pattern example for `runConsensus()`. It demonstrates two variations:

1. **Basic**: one proposer, two judges, `mode: 'refute'`, `quorum: 1`, `onDissent: 'revise'` — uses the default judge prompt.
2. **Custom `judgePrompt`**: same setup but each judge receives a different framing (security lens vs. operational-cost lens) via a `judgePrompt` function, showing how per-judge angles differ from the default skeptic framing.

Also adds one row to `examples/README.md` patterns table.

## Why

PR #280 landed `runConsensus()` and the `verify` hook and explicitly listed adding `examples/patterns/consensus.ts` as a good follow-up. No PR existed for it.

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` passes (pre-existing Windows symlink failures in `fs-walk.test.ts` / `built-in-tools.test.ts` are unrelated)
- [x] Example follows the comment-header + `../../src/index.js` import pattern used by all other pattern examples

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
-analyze-*.md
-tracking.md
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
+analyze-*.md
+tracking.md
--- a/examples/README.md
+++ b/examples/README.md
@@ -58,6 +58,7 @@ Reusable shapes for common multi-agent problems.
 | [`patterns/cost-tiered-pipeline`](patterns/cost-tiered-pipeline.ts) | Run the same four-stage pipeline twice to compare flagship vs tiered model cost. |
 | [`patterns/agent-handoff`](patterns/agent-handoff.ts) | Synchronous sub-agent delegation via `delegate_to_agent`. |
 | [`patterns/plan-replay`](patterns/plan-replay.ts) | Pin a coordinator plan with `createPlanArtifact`, then replay it with `runFromPlan`, no coordinator re-run. |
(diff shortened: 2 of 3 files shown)


JackChen-me:
Example's solid. Two fixes before merge:

Drop the `.gitignore` scratch entries (`analyze-*.md`, `tracking.md`). Same as #301, looks like your branch base carries them into every PR. Kill them at the source so they stop riding along.

Variation 2's `mode: 'refute'` does nothing: once `judgePrompt` is set, `buildJudgePrompt` ignores `mode`. Drop it. (`mode: 'lens'` is the built-in version of the per-judge angles you're hand-rolling, if you want that path.)

nuthalapativarun:
@JackChen-me Thanks for the review — both fixed in f1e211b:

1. **`.gitignore` scratch entries**: removed `analyze-*.md` and `tracking.md` from `.gitignore` in this branch. These were introduced at the shared branch base (`a9f2af0`, the `.gitignore` merge-conflict resolution commit) that #297/#301/#302 are all built on top of — removing them directly in this commit so this branch stops carrying them forward.
2. **`mode: 'refute'` no-op**: removed it from Variation 2. Confirmed in `buildJudgePrompt` that `judgePrompt !== undefined` short-circuits before `mode` is ever consulted, so it was dead config. Variation 1 still demonstrates the `mode: 'refute'` default path.

`npm run lint` and `tests/consensus.test.ts` (20 tests) pass. Pushed to `feat/consensus-pattern-example`.

(Note: I don't see this as an inline review thread on GitHub — it looks like a top-level PR comment — so there's no review thread to resolve via the API; replying here to confirm the fix instead.)

PR #298, opened 2026-06-16
docs(providers): add Azure OpenAI setup guide

nuthalapativarun:
## What

Adds `docs/providers/azure-openai.md` — a setup guide for Azure OpenAI following the pattern established by `docs/providers/minimax.md`.

Covers:
- Required and optional environment variables (`AZURE_OPENAI_API_KEY`, `AZURE_OPENAI_ENDPOINT`, `AZURE_OPENAI_API_VERSION`, `AZURE_OPENAI_DEPLOYMENT`)
- Agent config with the deployment-name-vs-model-name clarification
- Inline full example
- Link to the runnable `examples/providers/azure-openai.ts`
- Model deployment name table

## Why

`docs/providers/minimax.md` established a per-provider doc pattern. Azure OpenAI, Gemini, and Qiniu still have no dedicated pages. Azure OpenAI is one of the most commonly used providers and the deployment-name vs model-name distinction is a frequent source of confusion not captured elsewhere.

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
+analyze-*.md
+tracking.md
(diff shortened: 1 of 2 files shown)


JackChen-me:

Worth keeping one bit though: the deployment-name vs model-name distinction isn't really in `providers.md`. A line on that in the Azure row would be a useful PR.

PR #299, opened 2026-06-16
docs(providers): add Gemini setup guide

nuthalapativarun:
## What

Adds `docs/providers/gemini.md` — a setup guide for Google Gemini following the `minimax.md` pattern.

Covers:
- `@google/genai` peer dependency install step (Gemini is the only first-class adapter that isn't OpenAI-compatible)
- `GEMINI_API_KEY` setup with link to Google AI Studio
- Agent config and inline full example
- Link to `examples/providers/gemini.ts`
- Model table (`gemini-2.5-flash` / `gemini-2.5-pro`) with notes
- Brief note on native thinking block / `thoughtSignature` support and cross-provider reasoning opt-in

## Why

`docs/providers/minimax.md` established a per-provider doc pattern. Gemini is the second most-referenced provider in the README after Anthropic and has unique setup requirements (non-OpenAI-compatible SDK, optional peer dep install step) that aren't captured in the main `docs/providers.md` table row.

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
+analyze-*.md
+tracking.md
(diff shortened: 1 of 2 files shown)


JackChen-me:

PR #300, opened 2026-06-16
docs(providers): add Qiniu setup guide

nuthalapativarun:
## What

Adds `docs/providers/qiniu.md` — a setup guide for Qiniu AI following the `minimax.md` pattern.

Covers:
- `QINIU_API_KEY` setup with link to the Qiniu developer console
- Agent config (`provider: 'qiniu'`) with note that the base URL is pre-configured
- Inline full example
- Link to `examples/providers/qiniu.ts`
- Representative model table (DeepSeek V3/R1, Qwen, Llama) with link to the Qiniu catalog for the full list

## Why

`docs/providers/minimax.md` established a per-provider doc pattern. Qiniu has no dedicated page despite being in the provider table and having a runnable example. Its main distinguishing features — aggregating multiple model families behind one endpoint and the `deepseek-r1` thinking-mode option — aren't visible from the table row alone.

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
+analyze-*.md
+tracking.md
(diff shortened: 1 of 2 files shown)


JackChen-me:

PR #301, opened 2026-06-16
feat(orchestrator): wire verify hook into runTeam coordinator tasks

nuthalapativarun:
## What

Closes the gap noted in PR #280: *"Coordinator-generated tasks in `runTeam` can't yet opt in (the coordinator emits task JSON without a `verify` field)."*

**New API**: `RunTeamOptions.verifyJudges`

```ts
const result = await oma.runTeam(team, goal, {
  verifyJudges: [judgeAgentConfig],
})
```

When `verifyJudges` is set, the coordinator prompt gains a `"verify"` field description. The coordinator can opt individual tasks into consensus verification by emitting:
- `"verify": true` — apply with all defaults
- `"verify": { "mode": "refute", "quorum": 1 }` — partial config

`loadSpecsIntoQueue` merges the coordinator's partial spec with the caller-supplied judge roster to form a full `ConsensusVerifyOptions`. Explicit `runTasks` tasks with their own `ConsensusVerifyOptions` (including judges) are unaffected.

**`TaskExecutionRecord.verify`**: also added so `planOnly` snapshots surface the verify config and it can be inspected.

## Why

PR #280 explicitly listed `runTeam` verify integration as a follow-up. The gap was architectural: the coordinator can't emit full `AgentConfig` objects for judges in its JSON output, so it had no way to opt tasks in even if the user passed verify config.

The solution is a judge roster supplied at the call site (`verifyJudges`) that the coordinator JSON references implicitly via `"verify": true` or a partial spec. This keeps coordinator JSON clean while giving users full control over which judges are used.

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` passes (5 new tests in `tests/runteam-verify.test.ts`; pre-existing Windows symlink failures in `fs-walk.test.ts` and `built-in-tools.test.ts` are unrelated)
- [x] Added tests for: `verify: true`, partial spec merge, no-verify-without-judges, tasks without verify field, invalid verify values
- [x] Additive change — existing `runTeam` and `runTasks` callers unaffected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
-analyze-*.md
-tracking.md
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
+analyze-*.md
+tracking.md
--- a/src/orchestrator/orchestrator.ts
+++ b/src/orchestrator/orchestrator.ts
@@ -371,6 +371,19 @@ export async function executeWithRetry(
 // Parsed task spec (result of coordinator decomposition)
 // ---------------------------------------------------------------------------
 
+/**
+ * Partial verify config that the coordinator can emit in task JSON.
+ * Contains only fields that are safe to include in LLM output — `judges`
+ * (full AgentConfig objects) are always supplied by the caller via
+ * {@link RunTeamOptions.verifyJudges} and are never in coordinator JSON.
+ */
+interface CoordinatorVerifySpec {
+  readonly mode?: 'refute' | 'lens'
+  readonly quorum?: number
+  readonly maxRounds?: number
+  readonly onDissent?: 'revise' | 'reject' | 'keep'
+}
+
 interface ParsedTaskSpec {
   title: string
   description: string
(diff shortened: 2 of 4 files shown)


JackChen-me:
Tested locally, lint + full suite pass. One fix though: the `verify` JSDoc on `TaskExecutionRecord` says "inspection and replay", but `createPlanArtifact` drops it (`PlanTaskArtifact` has no `verify` field), so a plan round-trip loses the config. Just drop "and replay" from the doc and the PR description. Also pull the `.gitignore` scratch entries (`analyze-*.md`, `tracking.md`) out of the diff.

nuthalapativarun:
@JackChen-me Thanks for testing — fixed in 95dc0ff:

1. **`verify` JSDoc**: dropped "and replay" from the `TaskExecutionRecord.verify` doc comment in `src/types.ts` — it now just says "Populated in `planOnly` results for inspection." Confirmed `createPlanArtifact` only carries `id/title/description/assignee/dependsOn/memoryScope/maxRetries/retryDelayMs/retryBackoff` into `PlanTaskArtifact`, so `verify` does in fact drop out of a plan round-trip today. Also updated the PR description on GitHub to remove the same "inspected and replayed" phrasing.
2. **`.gitignore` scratch entries**: removed `analyze-*.md` and `tracking.md` from `.gitignore` (same root cause as #297 — they were added at the shared branch base, fixed directly in this branch's diff).

`npm run lint` and `tests/runteam-verify.test.ts` (5 tests) pass. Pushed to `feat/runteam-verify-hook`.

(This looks like a top-level PR comment rather than an inline review thread, so there's no thread to resolve via the GraphQL API — replying here to confirm the fix.)

PR #302, opened 2026-06-16
feat(bedrock): implement native reasoning round-trip via signature (#223)

### LINKED ISSUE #223: [Proposal] Cross-provider reasoning fallback (`<thinking>` text downgrade)

nuthalapativarun:
## Summary

Completes the Bedrock extended-thinking multi-turn round-trip by wiring both legs of the `reasoningContent.reasoningText.signature` protocol:

- **Inbound `chat()`** — `fromBedrockContentBlock` now extracts `signature` from `reasoningText.signature` and `redactedContent` bytes (converted to base64 `redactedData`) into the `ReasoningBlock` IR.
- **Inbound `stream()`** — the streaming loop now handles the `reasoningContent.signature` delta variant (a separate `SignatureMember` event that follows the text deltas); the signature is attached to the accumulated `ReasoningBlock` at `contentBlockStop`.
- **Outbound `toBedrockContentBlock`** — `'reasoning'` blocks with `provenance === 'bedrock'` are now echoed natively: blocks with a `signature` emit `{ reasoningContent: { reasoningText: { text, signature } } }`; blocks with `redactedData` emit `{ reasoningContent: { redactedContent: <bytes> } }`. Unsigned or foreign-provenance blocks fall back to the existing `<thinking>` text path (or are dropped silently when `preserveReasoningAsText` is off).
- **Capabilities** — `echoesReasoning` is promoted from `'never'` to `'own-issued'`, unblocking the cross-provider reasoning-fallback logic in `reasoning-fallback.ts`.

## Tests

9 new tests added to `tests/bedrock-adapter.test.ts` under `reasoning round-trip (#223 Phase 3)`:

- `chat()`: extracts `signature` from `reasoningText` into `ReasoningBlock`
- `chat()`: extracts `redactedContent` bytes into `ReasoningBlock.redactedData` (base64)
- Outbound: bedrock-provenance block with `signature` echoes natively (no `<thinking>`)
- Outbound: native echo fires even when `preserveReasoningAsText` is off
- Outbound: bedrock-provenance block without signature still falls back to text path
- Outbound: bedrock-provenance `redactedData` block echoes via `redactedContent`
- `stream()`: signature delta accumulates into `ReasoningBlock` in done payload
- `stream()`: reasoning block without signature delta has no `signature` field
- `capabilities.echoesReasoning` is `'own-issued'`

`tests/llm-adapter-capabilities.test.ts` updated to expect `'own-issued'` for the bedrock adapter.

All 52 tests in the two affected files pass; the full suite passes modulo pre-existing Windows-environment failures in `fs-walk.test.ts` and `built-in-tools.test.ts` (bash/symlink, unrelated to this change).

Closes #223

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
-analyze-*.md
-tracking.md
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ oma-dashboards/
 .claude/
 .npmrc
 .env
+analyze-*.md
+tracking.md
--- a/docs/context-management.md
+++ b/docs/context-management.md
@@ -92,7 +92,8 @@ When the fallback fires:
 | Matches target (`'deepseek'` → DeepSeek) in a tool-calling conversation | `'tool-use-only'` | Native `reasoning_content` echo (per DeepSeek V4 spec, see #251) |
 | Matches target but no `tool_use` anywhere in history | `'tool-use-only'` | Dropped (DeepSeek spec ignores non-tool reasoning) |
 | Foreign (e.g. `'openai'` → DeepSeek) | `'tool-use-only'` | Text fallback |
-| Any | `'never'` (OpenAI, Azure, Copilot, Bedrock, AI SDK, etc.) | Text fallback |
+| Matches target (`'bedrock'` → Bedrock) and has signature or is redacted | `'own-issued'` | Native echo via `reasoningContent.{reasoningText,redactedContent}` (see #223) |
+| Any | `'never'` (OpenAI, Azure, Copilot, AI SDK, etc.) | Text fallback |
 
 Redacted reasoning (Anthropic safety-filtered) emits the placeholder `<thinking>[redacted]</thinking>` to signal that reasoning occurred without leaking the opaque payload.
 
@@ -100,6 +101,6 @@ Redacted reasoning (Anthropic safety-filtered) emits the placeholder `<thinking>
 - Disabled by default to avoid silently inflating prompt tokens.
 - Default-on truncation (`compressReasoningText`) is mandatory for safety on long chain-of-thought; disable only when debugging.
 - Some local OpenAI-compatible models may echo `<thinking>` text back into their assistant response, which can trip the loop detector. See `examples/patterns/cross-provider-reasoning.ts` for the failure mode and mitigations.
-- Bedrock currently has `capabilities.echoesReasoning === 'never'` until a follow-up restores its signature round-trip on both inbound extraction and outbound serialization; until then, opt-in always falls back to text for Bedrock targets.
+- Bedrock has `capabilities.echoesReasoning === 'own-issued'`: signed reasoning blocks (`reasoningContent.reasoningText.signature`) and redacted blocks (`reasoningContent.redactedContent`) round-trip natively on both `chat()` and `stream()`, in both inbound extraction and outbound serialization (see #223).
 - `'tool-use-only'` (DeepSeek V4) is the only capability where same-provider echo works **without** the user opting into `preserveReasoningAsText` — it's forced on internally because the DeepSeek API requires it.
 
--- a/src/llm/bedrock.ts
+++ b/src/llm/bedrock.ts
@@ -79,13 +79,12 @@ function base64ToUint8Array(b64: string): Uint8Array {
  * Convert a single framework {@link ContentBlock} into a Bedrock
  * {@link BedrockContentBlock} for the messages array.
  *
- * Reasoning blocks are converted via the shared cross-provider fallback
- * (see #223): dropped silently by default; emitted as a standalone text
- * block wrapped in `<thinking>...</thinking>` when
- * {@link LLMChatOptions.preserveReasoningAsText} is `true`. Native echo
- * (which would require writing the signature into Bedrock's
- * `reasoningContent.reasoningText.signature` shape) is still pending —
- * see {@link BedrockAdapter.capabilities} for the gating rationale.
+ * Reasoning blocks with Bedrock provenance and a signature are echoed
+ * natively via `reasoningContent.reasoningText.{text,signature}`. Redacted
+ * blocks with Bedrock provenance are echoed via `reasoningContent.redactedContent`.
+ * All other reasoning blocks (foreign provenance, no signature) fall back to
+ * the cross-provider `<thinking>` text path when `preserveReasoningAsText` is
+ * set, or are dropped silently.
  */
 function toBedrockContentBlock(
   block: ContentBlock,
@@ -124,9 +123,18 @@ function toBedrockContentBlock(
     }
 
     case 'reasoning': {
-      // Capability is `'never'` (see BedrockAdapter.capabilities) — all
-      // reasoning falls back to text when the user opts in, regardless of
-      // provenance. When opt-in is off, drop silently (today's default).
+      if (block.provenance === 'bedrock') {
+        if (block.redactedData !== undefined) {
+          return {
+            reasoningContent: { redactedContent: base64ToUint8Array(block.redactedData) },
+          } as unknown as BedrockContentBlock
+        }
+        if (block.signature !== undefined) {
+          return {
+            reasoningContent: { reasoningText: { text: block.text, signature: block.signature } },
+          } as unknown as BedrockContentBlock
+        }
+      }
       if (outboundOptions?.preserveReasoningAsText !== true) return null
       const maxChars = resolveReasoningOutboundMaxChars(outboundOptions)
       const text = maxChars === undefined
@@ -204,10 +212,20 @@ function fromBedrockContentBlock(block: BedrockContentBlock): ContentBlock | nul
     return toolUse
   }
   if (block.reasoningContent !== undefined) {
-    const r = block.reasoningContent as { reasoningText?: { text?: string } }
+    const r = block.reasoningContent as { reasoningText?: { text?: string; signature?: string }; redactedContent?: Uint8Array }
+    if (r.redactedContent !== undefined) {
+      const reasoning: ReasoningBlock = {
+        type: 'reasoning',
+        text: '',
+        redactedData: Buffer.from(r.redactedContent).toString('base64'),
+        provenance: 'bedrock',
+      }
+      return reasoning
+    }
     const reasoning: ReasoningBlock = {
       type: 'reasoning',
       text: r.reasoningText?.text ?? '',
+      ...(r.reasoningText?.signature !== undefined ? { signature: r.reasoningText.signature } : {}),
       provenance: 'bedrock',
     }
     return reasoning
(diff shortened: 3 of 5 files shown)


JackChen-me:
stream() handles the signature delta but not redactedContent (ReasoningContentBlockDelta is a union of all three), so a redacted thinking block over streaming gets dropped even though chat() handles it. Narrow case, but since this closes #223 I'd rather fix it here than leave the round-trip asymmetric. Otherwise I'll open a follow-up.

Also two stale doc lines once this merges: docs/context-management.md:95 still lists Bedrock under 'never', and :103 says it's pending exactly this follow-up. Fix both in the PR.

nuthalapativarun:
@JackChen-me Good catch — fixed in c101d90:

1. **`redactedContent` dropped in `stream()`**: `ReasoningContentBlockDelta` deltas are now buffered for all three variants (`text`, `signature`, `redactedContent`) keyed by content block index, mirroring what `chat()` already does via `fromBedrockContentBlock`. On `contentBlockStop` (and the end-of-stream safety net), a buffered `redactedContent: Uint8Array` is flushed into `ReasoningBlock.redactedData` as base64, same as the non-streaming path. Added a test (`stream(): accumulates redactedContent delta into ReasoningBlock.redactedData (base64) in done payload`) mirroring the existing signature-delta test.
2. **Stale docs**: updated `docs/context-management.md` — the capability table now lists Bedrock (`'own-issued'`) with native round-trip via `reasoningContent.{reasoningText,redactedContent}` instead of grouping it under `'never'`, and the note that previously said it was "pending exactly this follow-up" now describes the round-trip as implemented on both `chat()` and `stream()` (closes #223).

`npm run lint` and `tests/bedrock-adapter.test.ts` (37 tests, including the new one) pass. Pushed to `feat/bedrock-reasoning-round-trip`.

(This is a top-level PR comment rather than an inline review thread on GitHub, so there's no thread to resolve via the GraphQL API — replying here to confirm the fix.)

PR #335, opened 2026-06-28
Add run.pay integration example (MemoryStore backend)

PalabreX:
Adds `examples/integrations/with-run.pay/` — a `RunPayMemoryStore` class
implementing OMA's `MemoryStore` interface, backed by run.pay's
pay-per-call Memory Store service. Durable shared memory across restarts
without self-hosting Redis/Postgres.

Also includes generic `list_runpay_services` / `call_runpay_service` tools
for reaching the rest of run.pay's infra (Task Scheduler, Idempotency
Lock, Cost Calculator, Consensus Aggregator) or any vendor service on
the marketplace.

Reciprocal reference added on run.pay's docs:
https://getrunpay.com/docs.html (Memory Store section)

### DIFF
--- a/examples/integrations/with-run.pay/README.md
+++ b/examples/integrations/with-run.pay/README.md
@@ -1,0 +1,83 @@
+# with-run.pay
+
+[run.pay](https://getrunpay.com) is a Stripe-native marketplace offering
+pay-per-call agent infrastructure: Memory Store, Task Scheduler,
+Idempotency Lock, Cost Calculator, and Consensus Aggregator — plus
+third-party vendor services. No self-hosting, no accounts, billed per call.
+
+This example shows two ways to plug run.pay into OMA:
+
+1. **`RunPayMemoryStore`** — a `MemoryStore` implementation backed by
+   run.pay's Memory Store service, passed as `sharedMemoryStore` on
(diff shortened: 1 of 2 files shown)


JackChen-me:
This won't run as-is. Three breaks against the current API:

`oma.registerTool(...)` isn't a method on `OpenMultiAgent`, so it throws at startup. Tools reach an agent via `AgentConfig.customTools` (see `examples/integrations/with-engram/team-research.ts`), and the `tools: ['list_runpay_services', ...]` name refs won't resolve without that.

`RunPayMemoryStore` doesn't match `MemoryStore`: `get()` must return `Promise<MemoryEntry | null>` (with `.value` / `.createdAt`), not `string | undefined`, since `SharedMemory` reads `.value` off it. And `list()` is `list(): Promise<MemoryEntry[]>`, no `prefix` param, entries not key strings. Copy `examples/integrations/with-tencentdb-memory/tdam-store.ts`.

examples are excluded from `tsc`, so green CI just means the build didn't break. Run it once against a live wallet first, the `MemoryEntry` mismatch surfaces immediately.

JackChen-me:
Does this PR still work for you? Please let me know. Otherwise, I'll close it in a few days.

PR #345, opened 2026-07-01
feat: add run-level metrics rollup on TeamRunResult

### LINKED ISSUE #341: [Feature] Run-level metrics rollup on TeamRunResult (tokens / retries / errors / latency)
## Context
Per-span and per-task metrics already exist: `TaskTrace.retries`, per-call token usage, per-task `TaskExecutionMetrics` (durationMs / tokenUsage / toolCalls), and `TeamRunResult.totalTokenUsage`.

## Gap
There is no run-level rollup of these into a machine-readable summary. A caller wanting total tokens, total retries, error rate, or latency aggregates for a run has to fold the raw per-span/per-task fields themselves. The post-run dashboard also does not render run-level totals.

## Possible direction
Compute a run-level metrics summary (tokens, retry count, error/failure count, task latency aggregates) from data already collected, and expose it on `TeamRunResult`. Cheap because it needs aggregation only, no new instrumentation.

lesbass:
## Summary

Adds run-level metrics aggregation to `TeamRunResult`, computing total tokens, retries, error counts, and task latency aggregates from per-task data already collected during orchestration. The dashboard also gains a metrics summary bar.

Closes #341.

## Changes

- **`types.ts`**: Add `RunMetrics` interface (totalTokens, totalRetries, errorCount, failureCount, completedCount, duration aggregates). Add `retries` field to `TaskExecutionMetrics`. Add optional `metrics` field to `TeamRunResult`.
- **`orchestrator.ts`**: Add `computeRunMetrics()` helper that aggregates from task records. Populate `retries` in per-task metrics during execution. Call `computeRunMetrics` in `buildTeamRunResult`.
- **`dashboard/render-team-run-dashboard.ts`**: Render a compact run-level metrics summary bar (tokens, retries, errors, completed/total, duration) from `result.metrics`.
- **`index.ts`**: Re-export `RunMetrics` type.

## Verification

- All 1066 tests pass (70 test files)
- No new TypeScript errors introduced
- Dashboard tests pass with the new metrics bar rendering

### DIFF
--- a/packages/core/src/dashboard/render-team-run-dashboard.ts
+++ b/packages/core/src/dashboard/render-team-run-dashboard.ts
@@ -23,6 +23,7 @@ export function renderTeamRunDashboard(result: TeamRunResult): string {
     generatedAt,
     goal: result.goal ?? '',
     tasks,
+    metrics: result.metrics ?? null,
     layout: {
       positions: serializedPositions,
       width: layout.width,
@@ -189,7 +190,10 @@ export function renderTeamRunDashboard(result: TeamRunResult): string {
         .bg-surface-container-low { background: var(--surface-container-low); }
         .bg-surface-container-lowest { background: var(--surface-container-lowest); }
         .bg-surface-variant { background: var(--surface-variant); }
+        .rounded { border-radius: 0.25rem; }
         .text-primary { color: var(--primary); }
+        .text-secondary { color: var(--secondary); }
+        .text-tertiary { color: var(--tertiary); }
         .text-secondary { color: var(--secondary); }
         .text-tertiary { color: var(--tertiary); }
         .text-error { color: var(--error); }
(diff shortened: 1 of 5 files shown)


JackChen-me:
The `TeamRunResult` rollup is the core of #341 and it's basically there. Suggest landing just that here (one bug + a test below) and taking the dashboard bar as a separate PR.

`avgTaskDurationMs` comes out inflated when a run has failures. `taskMetrics.set` runs before the `result.success` branch (`orchestrator.ts:1052`), so a task that ran and then failed still records metrics, and `computeRunMetrics` folds its `durationMs` into `totalDurationMs` with every other timed task (`:257`) while dividing by `completedCount` (`:275`). A run with a slow failed task can then report an average above `maxTaskDurationMs`. Divide by the timed-task count, or only sum completed durations. No coverage on `computeRunMetrics` yet, so a test over a run with one failed task would lock it in.

The dashboard bar is why the split: it makes the DAG disappear on desktop. `runMetricsBar` is the first flex child of `<main>` with `w-full`, so at ≥1024px (main is `flex-row`) it eats the full row and collapses the `flex-1` viewport to zero (rendered it at 1300px to check). It also uses classes that aren't in the inline `<style>` (`bg-surface-container-high`, `gap-6`, `flex-wrap`), so no background and the fields run together. Both fixable, just more than the rollup should wait on.

lesbass:
Thanks for the review. Addressed all three points:

1. **Dashboard bar removed** — reverted `render-team-run-dashboard.ts` changes. Will submit the bar as a separate PR.
2. **Fixed `avgTaskDurationMs` inflation** — `computeRunMetrics` now only accumulates duration metrics from completed tasks (`task.status === 'completed'`). Token/retry totals still aggregate from all tasks since they're run-level sums.
3. **Added test** — new test `computes run-level metrics from mixed completed and failed tasks` exercises `computeRunMetrics` with one failed + one completed task, verifying `failureCount`, `completedCount`, `avgTaskDurationMs`, etc.

All 1052 tests pass (51 files). Ready for re-review.

PR #357, opened 2026-07-06
fix: kill full process tree on Windows bash timeout

Bobuyoucrypto:
## What

Fixes the built-in `bash` tool on Windows: a timed-out command now kills its full process tree and returns promptly, instead of leaving background children alive and hanging until they exit naturally. Also makes the test suite pass on Windows (was 8 failures; now 0 failures, 6 environment-gated skips).

## Why

On Windows, `killProcessTree` fell back to `child.kill('SIGKILL')`, which terminates only the direct bash process. Two bugs followed:

1. **Background children survive the timeout.** Git Bash (MSYS) emulates `fork()` through short-lived intermediate processes, so a descendant's recorded Win32 parent PID often points at a process that no longer exists. No parent-chain walk (including `taskkill /T`) can reach such orphans.
2. **The tool call hangs.** Any surviving descendant inherits the stdio pipes, and the implementation waited for the `close` event (process exit AND stdio drained), so the tool blocked until the orphan exited on its own. Reproduced with `(sleep 0.4; echo alive) & sleep 5` and a 100 ms timeout: the call returned after ~5 s.

Fix, in three parts:

- Use `taskkill /pid <pid> /T /F` on win32 — kills everything still reachable through live parent links (in practice this wins the race against short-lived background children).
- After a forced kill, settle on `exit` instead of `close`, so an unreachable MSYS straggler holding the pipes cannot delay the result.
- Report exit codes 124 (timeout) / 130 (abort) authoritatively: the OS-reported code after a forced kill is platform-dependent (`null` after SIGKILL on POSIX, `1` from taskkill on Windows), which broke the documented 124 contract.

Test-only changes for Windows compatibility:

- New `tests/helpers/symlink-support.ts` probe; the six symlink sandbox-escape tests (`built-in-tools`, `fs-walk`) skip via `it.skipIf` when the environment cannot create symlinks (Windows without Developer Mode raises `EPERM`). They still run everywhere else, including Linux CI.
- The bash `pwd` cwd test now asserts on the unique `mkdtemp` basename, since Git Bash prints the MSYS translation (`/tmp/oma-test-X`) of the native Windows path.

Verified on Windows 11, Node 26: `npm run lint && npm test` → 1108 passed, 6 skipped, 0 failed (three consecutive runs of the timing-sensitive kill test, all stable). POSIX behavior is unchanged: the process-group SIGKILL path is untouched, and the `exit`-settling path only activates after a forced kill, where `exit` and `close` coincide.

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

### DIFF
--- a/packages/core/src/tool/built-in/bash.ts
+++ b/packages/core/src/tool/built-in/bash.ts
@@ -149,9 +149,31 @@ function runCommand(
       signal.addEventListener('abort', onAbort, { once: true })
     }
 
+    // `close` (process exited AND stdio drained) is the normal completion
+    // path. After a forced kill we settle on `exit` instead: on Windows,
+    // MSYS bash's fork emulation can leave descendants that taskkill cannot
+    // reach (their recorded parent PID is a dead intermediate process), and
+    // any such straggler would hold the stdio pipes open and delay `close`
+    // until it exits naturally.
+    //
+    // When we killed the process ourselves, report the conventional exit
+    // codes (124 timeout / 130 abort) authoritatively: the code the OS
+    // reports for a forced termination is platform-dependent (`null` after
+    // SIGKILL on POSIX, `1` after taskkill on Windows).
+    const resolveExitCode = (code: number | null): number => {
+      if (timedOut) return 124
+      if (aborted) return 130
+      return code ?? 1
+    }
+
     child.on('close', (code: number | null) => {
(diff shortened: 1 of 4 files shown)


JackChen-me (approved):
Thanks, LGTM

JackChen-me:
What are you running OMA on Windows for? Would be useful to know.

PR #358, opened 2026-07-08
Add orchestrator cost budget enforcement

LambIessz:
## Summary
- Add `maxCostBudget` and `estimateCost` to orchestrator config for user-defined cost caps.
- Pass effective model/provider/phase/taskId into the estimator so callers can price per model.
- Enforce estimated cost budgets across `runAgent`, `runTasks`, `runTeam`, consensus, and synthesis paths, mirroring token-budget boundaries.

## Test Plan
- [x] npm test --workspace @open-multi-agent/core -- token-budget.test.ts
- [x] npm run lint --workspace @open-multi-agent/core
- [x] npm run build --workspace @open-multi-agent/core

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -194,6 +194,23 @@ const orchestrator = new OpenMultiAgent({
 })
 ```
 
+**Cap estimated cost.** Keep the price table in your app and provide an estimator. OMA passes the effective `model`, `provider`, phase, and `taskId` (when available) so you can price per model. The cap is checked at the same turn/task boundaries as `maxTokenBudget`, so a run can overshoot by up to one model turn; it is not a cent-exact stop.
+
+```ts
+const prices = {
+  'gpt-4.1-mini': { input: 0.40, output: 1.60 }, // USD per 1M tokens
+}
+
+const orchestrator = new OpenMultiAgent({
+  maxCostBudget: 0.25,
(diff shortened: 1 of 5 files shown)


JackChen-me:
Mechanics look right, token totals match before/after. Two things, both in `orchestrator.ts`:

`maxCostBudget` silently does nothing without `estimateCost`, and nothing warns you. Constructor takes both as-is (`:1771`), cost check bails when the estimator's undefined (`:360`). Worst failure mode for a spend breaker: you think you're capped and you aren't. Worth validating the pairing in the constructor.

Smaller: a bad estimator throws (`:319`), uncaught in the worker dispatch (`:1238`), so it rejects the whole run instead of degrading like a task failure. A price-table slip returning NaN is enough. Worth a guard + a test.

LambIessz:
Fixed in `79994bb`.

- Added constructor validation so `maxCostBudget` cannot be configured without `estimateCost`.
- Guarded worker cost accounting so an invalid/throwing estimator marks the current task failed instead of rejecting the whole run.
- Added regression tests for both cases.

Validation:
- `npm test --workspace @open-multi-agent/core -- token-budget.test.ts`
- `npm run lint --workspace @open-multi-agent/core`
- `npm run build --workspace @open-multi-agent/core`

Issue #204, opened 2026-05-03
External agent orchestration: CLI tools and business systems as team members


JackChen-me:
## What

OMA orchestrates LLM-backed agents today. This issue tracks extending OMA to also orchestrate external agents that run as local processes: coding agent CLIs (Claude Code, Codex, Cursor Agent, Gemini CLI) now, and business-system CLIs and APIs (Lark, GitHub, Linear, Stripe, etc.) longer term.

The motivating use case is a hybrid team: an LLM-backed planner decomposes the goal, a CliAgent spawns Claude Code in the project directory to write code, an LLM-backed reviewer audits the diff. All three run in one OMA task DAG with unified token accounting, shared memory, and error isolation.

## Why now

Agent Client Protocol (ACP) standardized the wire format for external agent integration with 40+ implementing agents (Claude Code, Codex, Cursor, Gemini CLI, Cline, Goose, JetBrains Junie, and more). Editor adoption across Zed, JetBrains, Neovim, and Emacs makes external agents a real category. Existing multi-CLI orchestration projects (multica, composio agent-orchestrator, awslabs/cli-agent-orchestrator, open-design) confirm demand exists, but none are library-shaped and embeddable into your own backend.

## How to contribute

Open to RFCs on the abstraction (CliAgent class shape, AgentBackend interface, ACP compatibility), PRs adding adapters for specific CLIs, or use cases that should drive the v1 surface. Comment here with what angle you want to explore.

## Out of scope for this issue

Discussion stays on the abstraction and integration. Broader direction questions belong in separate discussions.

JackChen-me:
Competitive data point. Today Mastra shipped an ACP integration: https://mastra.ai/blog/introducing-agent-client-protocol. Their agent takes the ACP client/supervisor role and delegates coding to external harnesses (Claude, Codex, Cursor, Gemini CLI) running in isolated workspaces. Same "framework as supervisor driving external coding CLIs" direction this issue sketches, so the near-term coding-CLI slice isn't hypothetical anymore.

Where it differs: theirs is ACP-only. Here ACP is one compatibility target under the broader `CliAgent`/`AgentBackend` abstraction, with business-system CLIs/APIs in longer-term scope. The OMA-specific bet is putting a `CliAgent` into the existing task DAG alongside LLM agents (shared memory, unified token accounting, cascade error isolation), not just a supervisor/subagent handoff.

JackChen-me:
[...]

PR #378, opened 2026-07-16
feat(agent): add generic process backend

scarab-systems:
## What

Extends the external-agent v1 by adding a protocol-neutral `process` backend alongside the existing ACP backend. The ACP backend remains the richer protocol-specific path; the new backend runs a configured local command as an agent, supports stdin/argument/no prompt delivery, handles cancellation, redacts stderr on failures, and exposes a public `@open-multi-agent/core/process` subpath.

This also updates the external-agent docs, package entrypoint smoke checks, and adds a no-key runnable process-backend example.

## Why

Refs #204.

The existing ACP implementation is a strong v1 foundation for external agents that speak an agent protocol. This PR is intended as a follow-on extension to that work, not a replacement: it keeps ACP as the protocol-aware backend and adds a smaller generic backend for local CLIs, scripts, and adapters that only need prompt-in/stdout-out process semantics.

That lets simpler external process agents participate in the same `AgentBackend` flow as ACP agents, including `runTasks`, shared memory handoff, dependency ordering, cancellation, and cascade-on-failure behavior.

## Validation

- `npm run lint && npm test`
- `npm run build`
- `npm run test:scaffold`
- `npx tsx packages/core/examples/integrations/external-agent-process.ts`
- package entrypoint smoke for `dist/index.js`, `dist/acp.js`, `dist/process.js`, `dist/mcp.js`, and `dist/ai-sdk.js`

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

Disclosure: I used AI-assisted coding tools while preparing this PR. I reviewed the changes myself, tested them, and take responsibility for the implementation and any follow-up revisions needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Smoke-test that the built entry points import and the CLI runs
         working-directory: packages/core
         run: |
-          node -e "Promise.all([import('./dist/index.js'), import('./dist/mcp.js'), import('./dist/ai-sdk.js')]).then(() => console.log('entry points import OK')).catch((e) => { console.error(e); process.exit(1); })"
+          node -e "Promise.all([import('./dist/index.js'), import('./dist/acp.js'), import('./dist/process.js'), import('./dist/mcp.js'), import('./dist/ai-sdk.js')]).then(() => console.log('entry points import OK')).catch((e) => { console.error(e); process.exit(1); })"
           node dist/cli/oma.js help > /dev/null
           echo "CLI help runs OK"
       - name: Assert the create-oma-app template pins the current core version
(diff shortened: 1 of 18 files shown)


JackChen-me:
Direction makes sense; reusing `AgentBackend` is clean. Three blockers and one doc fix before merge:

**Breaking type change:** Changing `AgentBackendConfig` from the interface published in v1.10.0 to a union breaks `extends AgentBackendConfig` and direct `.permission` access (`TS2312`/`TS2339`). Please keep the existing ACP interface for compatibility and introduce a separately named union for `AgentConfig.backend`.

**Abort leaks descendants:** `child.kill()` without a detached process group only kills the direct child, so an aborted or timed-out agent can leave a live process tree. Extract and reuse the process-tree cleanup from `tool/built-in/bash.ts`.

**Early `stream()` exit orphans the child:** The generator's `finally` path removes listeners but never terminates the subprocess. Run the same process-tree cleanup there and wait for forced exit.

**Docs are still ACP-only:** The backend JSDoc says `process` requires the ACP SDK and `/acp` import, while the sandbox and permission warning doesn't cover `process`. The permission link should point to `AcpAgentBackendConfig.permission`.

The two lifecycle paths need regression coverage with the fixes.

scarab-systems:
I appreciate the careful review here. I pushed an update in `97a91d1` that addresses the compatibility, lifecycle, and docs points:

- Kept `AgentBackendConfig` as the ACP config interface for v1.10 compatibility, and introduced `ExternalAgentBackendConfig` for the backend union used by `AgentConfig.backend`.
- Extracted the existing process-tree cleanup into a shared helper and reused it for process backend aborts and early `stream()` exits.
- Added regression coverage for both lifecycle paths: aborted runs and consumers that stop reading a stream early.
- Updated the external-agent docs/JSDoc so `process` is no longer described as ACP-only, and the security guidance distinguishes ACP permission prompts from process backend command/cwd constraints.

I also reran the local checks (`npm run lint`, `npm test`, `npm run build`, plus the focused lifecycle/type checks), and the GitHub CI checks are green on the updated commit.

JackChen-me:
Original blockers are resolved. The parent-exits-before-break cleanup edge case can follow separately. LGTM.

PR #399, opened 2026-07-18
refactor(core): extract coordinator module (3/3)

JackChen-me:
## Summary

Refactor (3 of 3): extract the coordinator subsystem into `coordinator.ts` and
update the AGENTS.md layer map. Stacked on the PR2 branch.

`orchestrator.ts` drops from 2213 to 1683 lines (57% below the original 3953).
`coordinator.ts` holds goal-decomposition parsing, coordinator prompt
construction (system prompt, roster, output format, decomposition), the synthesis
pass, and task-spec loading.

## Move vs signature change

- The parse helpers (`parseTaskSpecs` and friends) are a pure move.
- The former private class methods become free functions. Two take the
  orchestrator config explicitly (`buildCoordinatorBaseConfig`,
  `runCoordinatorSynthesis`) instead of `this.config`; prompt string literals are
  byte-for-byte unchanged (only code indentation shifts as methods de-indent to
  module scope).
- One type follow-on: `runCoordinatorSynthesis` gains a `?? DEFAULT_MODEL` guard
  because `OrchestratorConfig.defaultModel` is optional where the resolved class
  field was not. It is dead at runtime (callers pass the resolved config) and
  mirrors the same guard already in `task-execution`'s `executeQueue`.

## Verification

- Parse block byte-identical to the original; all 87 single-quoted and 17
  backtick prompt-string literals identical.
- `npm run lint` clean, `npm test` green (93 files, 1385 tests).
- `git diff -- packages/core/tests` is empty.
- `dist/index.d.ts` is identical to the pre-split build.

## Stacking

Base branch: `refactor/orchestrator-split-execution` (PR2). Merge #397, then PR2,
then this.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,7 +44,7 @@ The framework's key feature. The coordinator receives goal + roster → emits a 
 
 | Layer | Files | Responsibility |
 |-------|-------|----------------|
-| Orchestrator | `orchestrator/orchestrator.ts`, `scheduler.ts` | Top-level API, task decomposition, retry/backoff, coordinator |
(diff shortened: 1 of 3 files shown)


JackChen-me:
Superseded by #401, rebuilt on `main` after #397 merged. The original branch was based on #397's pre-squash commit, which the updated `main` no longer contains; the replacement targets the current `main`.

PR #424, opened 2026-07-23
feat(core): add retryable route fallback

### LINKED ISSUE #350: [Feature] On-error model fallback: failover to a backup provider on retryable LLM errors
## Context

Two adjacent pieces already exist:
- Deterministic `modelRouting` (#271, #286/#293) picks a model/provider from a task's static shape, decided once before execution.
- Error-aware task retry (#337, PR #346) retries a failed task with `isRetryableError()` classification and jittered backoff, but re-runs the same model/provider.

## Gap

Neither switches provider when the primary one is down or persistently rate-limited. On a sustained outage or repeated 429s, a run just retries the same provider until it exhausts `maxRetries` or the token budget; it cannot degrade to a backup model. Users wanting "Claude to DeepSeek/GPT on failure" have to catch errors and swap adapters by hand.

## Possible direction

- Add an optional `fallback: ModelRouteConfig[]` to `ModelRoutingRule.route` (or a top-level fallback chain).
- Reuse `isRetryableError()`: when the active route fails with a retryable error past a threshold, advance to the next entry in the fallback list; terminal errors do not trigger failover.
- Off by default: with no `fallback` configured, behavior is unchanged.

## Caveat

`source:analysis`, no external user request. Motivation is maintainer dogfooding where a hosted provider intermittently degraded from certain deploy regions. Not validated as a paid-user pain point; filing to track the gap, not as market evidence.

LambIessz:
## Summary

- Add an ordered `fallback` chain to worker `modelRouting` routes.
- Advance the next task retry to the next configured route only for retryable provider errors.
- Price fallback attempts with the model/provider that actually handled them, while retaining aggregate cost estimation for single-route tasks.

## Motivation

Closes #350. Task retry already classifies retryable errors, but previously retried the same provider for every attempt. This lets callers configure a backup route for outages and persistent rate limits without mutating their Team or swapping adapters manually.

## Scope and impact

- Affected: `@open-multi-agent/core` and `docs/model-routing.md`.
- Public API: optional `ModelRouteConfig.fallback`; the behavior is opt-in and applies only to worker tasks, which already have task-level retry settings.
- Terminal errors such as 401 do not retry or fail over. When fallbacks are exhausted, later retries remain on the final route.
- Security and privacy: none. Each fallback uses only its own route overrides combined with the original worker configuration.
- Non-goals: coordinator, synthesis, short-circuit, and delegated phases do not gain a separate retry policy.

## Validation

- [x] `npm test --workspace @open-multi-agent/core -- orchestrator.test.ts` (55 tests)
- [x] `npm run lint --workspace @open-multi-agent/core`
- [x] `npm run build --workspace @open-multi-agent/core`
- [x] Added tests for retryable failover, ordered fallback exhaustion, terminal and unclassified errors, and per-route cost contexts.
- [x] No new runtime dependencies.

`npm test --workspace @open-multi-agent/core` was also run. Its local Windows-only failures are in independent file-store `fsync` permission and POSIX CLI subprocess suites; routing and retry coverage passes.
```

### DIFF
--- a/docs/model-routing.md
+++ b/docs/model-routing.md
@@ -52,6 +52,45 @@ The `route` is the override applied when the rule matches. Only `model` is requi
 | `baseURL` | Base URL for OpenAI-compatible or self-hosted endpoints. |
 | `apiKey` | API key for the matched call. |
 | `region` | AWS region, for Bedrock routes. |
+| `fallback` | Ordered backup routes for retryable worker-provider failures. |
+
+## Retryable provider fallback
+
+Worker routes can declare an ordered `fallback` list. The initial attempt uses
+the route itself. When a confirmed provider error is retryable, the next task
+retry moves to the next entry in the list. Once the list is exhausted, later
+retries keep using its final entry.
+
+```ts
+const modelRouting: ModelRoutingPolicy = {
+  rules: [{
+    match: { phase: 'worker' },
+    route: {
+      model: 'claude-sonnet-4-5',
+      provider: 'anthropic',
+      fallback: [{
+        model: 'gpt-5',
+        provider: 'openai',
+      }],
+    },
+  }],
+}
+
+await orchestrator.runTasks(team, [{
+  title: 'Write report',
+  description: 'Summarize the findings.',
+  assignee: 'writer',
+  maxRetries: 1,
+}], { modelRouting })
+```
+
+Fallback is opt-in and reuses each task's existing retry settings, so set
+`maxRetries` high enough to reach the backup entries you configure. Authentication,
+validation, hook, and other non-provider errors never advance the fallback chain.
+A failed task without a structured provider error continues to use its current
+route. Each fallback entry is applied to the worker's effective configuration like
+any other route. When a backup uses a different provider or endpoint, include its
+`provider`, `baseURL`, `apiKey`, and `region` overrides as appropriate.
 
 ## Which calls are routed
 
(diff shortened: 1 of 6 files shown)


JackChen-me:
Could you update this branch from `main` and restrict fallback to confirmed provider failures? Right now `beforeRun`/`afterRun` errors also advance to the next provider.

LambIessz:
Updated in 1f021ab.

- Rebased the branch onto the current `main`.
- Fallback now advances only when the failed run is classified as a retryable provider error (`errorInfo.kind === 'provider'`).
- `beforeRun`, `afterRun`, and unclassified/framework errors retry the current route instead.

Added regression coverage for `beforeRun` and `afterRun` errors carrying a 429-like status, and reran:

- `npm test --workspace @open-multi-agent/core -- orchestrator.test.ts task-retry.test.ts token-budget.test.ts`
- `npm run lint --workspace @open-multi-agent/core`
- `npm run build --workspace @open-multi-agent/core`

JackChen-me:
One blocker remains: with `onAgentStream` enabled, a provider 503 takes the thrown-error path and retries `primary-model` instead of advancing to the fallback. Please preserve provider classification through that path and add an `onAgentStream` regression test.

LambIessz:
Updated in fe4da52.

- Rebased the branch onto the current `main`.
- Streamed provider failures now preserve their normalized `errorInfo` through `onAgentStream`, so a retryable 503 advances to the next fallback route.
- `beforeRun`, `afterRun`, and other callback/framework failures remain classified separately and retry the current route.

Added an end-to-end `onAgentStream` regression test for primary-route 503 failover.

Validation:
- `npm test --workspace @open-multi-agent/core -- orchestrator.test.ts task-retry.test.ts token-budget.test.ts` (100 tests)
- `npm run lint --workspace @open-multi-agent/core`
- `npm run build --workspace @open-multi-agent/core`

PR #548, opened 2026-08-23
Add video input support for MiniMax-compatible chat

octo-patch:
Reason: MiniMax-M3 video input is not represented by ContentBlock or serialized by the OpenAI-compatible adapter.

This change adds a public VideoBlock with inline and URL sources, serializes video as an OpenAI-compatible video_url content part, and preserves video attachments when preparing context summaries. Adapters without video wire support reject the block explicitly. Provider documentation and focused adapter coverage are included.

Checks:
- npm run test --workspace @open-multi-agent/core -- tests/llm-adapters.test.ts
- npm run lint --workspace @open-multi-agent/core
- npm run build --workspace @open-multi-agent/core
- git diff --check


JackChen-me:
The PR states "Adapters without video wire support reject the block explicitly", but the `/ai-sdk` bridge is an exception: `toAISDKMessages` silently drops `video` blocks — a video-only user message disappears entirely (its user-parts loop only handles `text`/`image`, and AI SDK has no video part). Please throw an explicit error there to match the Anthropic/Bedrock/Gemini cases.

octo-patch:
Thanks for the review. I updated the AI SDK bridge to reject video content blocks explicitly, added a regression test, and pushed a new commit. The focused test, core lint, full core test suite, and core build all pass locally.

